Issue #34, opened 2020-06-15
ensure that kbld.k14s.io/images annotations is deterministically ordered


cppforlife:
to avoid unnecessary annotation changes

jbrunton:
@cppforlife: I have a bit of time and I'm looking to contribute to more OS projects. Any objections if I have a go at this? It might also pair up with #18 quite well.

cppforlife:
@jbrunton sounds good! it's all yours.

jbrunton:
@cppforlife: Thanks for the merging the PR! Can this issue be closed now too?

cppforlife:
@jbrunton i typically close all in-progress issues once they are included in a release. im planning to cut a new kbld version wed/thurs, so ill close it then.

cppforlife:
included in https://github.com/k14s/kbld/releases/tag/v0.24.0

PR #36, opened 2020-06-29
Order kbld.k14s.io/images annotation deterministically

jbrunton:
PR for #34. It sorts the images annotation by the image URL (I think the only reasonable field to sort on, since the Metas field can be nil).

The unit test is a little long in terms of loc but I think makes it easier for the reader to see how the annotation is being ordered.


cppforlife:
closing and reopening to pick up running GH test for PRs.

Issue #18, opened 2019-10-16
skip resolving images are referenced with their digest


dprotaso:
ie.

```
image: some/image@sha256:12312312...
```

Get's processed by `kbld` and resolves to itself. This also leaves unnecessary annotation on some objects`

ie.

```
kbld.k14s.io/images: |
  - Metas: null
    URL: some/image@sha256:12312312...
```

cppforlife:
thats fair. we should probably remove that.

Zebradil:
Processing resolved image might be useful to verify that the image is actually available.

pivotaljohn:
> Processing resolved image might be useful to verify that the image is actually available.

The guarantee that `kbld` is driving towards is ensuring that references are reliable — that they are digests (read: immutable) rather than tagged references.

The thinking goes that even if we were to check for the presence of the image, it doesn't prevent someone from removing it from the registry/repo moments later... it's not much of a guarantee. And in practice, if we've got digest-articulated coordinates, that URL is likely still legit.

In short, little gain for the cost of using a HTTP call to the registry (thinking of registry rate limiting).

Zebradil:
So far I haven't faced a situation when a digest was removed, but this is still possible.

In some workflows resolved digests are sitting in manifests for a long time (months maybe) and the probability of disappearing that particular digest exactly in the time frame between subsequent invocations of `kbld` and `kubectl` (or `kapp`) is rather small. I personally feel more comfortable when pipelines are falling earlier than later, so that I know about the issue before it gets to production and provokes alerts or unexpected behavior.

Also, there is a way to tell `kbld` to skip resolution of that particular image if it is not desired.

But, to be honest, I don't feel that this is required. I suppose that this scenario might be solved with `kbld` lock files. I just got the idea that it might be not a bug but a feature and thought that this also should be considered.

pivotaljohn:
Thank you for persisting in your perspective, @Zebradil.

One of our core design goals across the Carvel suite is to ensure the operator has confidence in the moves they make with the tools.

I can absolutely envision the situation where I've been handed a bill of materials and want to verify that these "materiel" are present. 👍🏼 

In fact, while `kbld` includes the annotation despite the fact that no real resolution occurred, it hasn't actually verified the presence of that image. 😮 

What do you think about including a flag that modifies this behavior to actually verify the presence of the image in the registry?  Something like `--verify-digest-refs` ?

Zebradil:
[...]

PR #39, opened 2020-07-24
Filter and de-dupe annotations

jbrunton:
This PR is a potential solution to #18. While implementing it I noticed a comment in one of the tests saying `TODO dedup same url images` which can be done at the same time, so I figured I'd attempt that too.

Note that digests already don't get _resolved_, they just get added to annotations with a null `Metas` field. So this change simply filters out those annotations (and any duplicates).

I'm creating this as a draft because:

1. I'm not sure how urgent either enhancement is, so I won't be offended if you decline one or other change (or both) – this was partly a learning exercise for me. Does it look useful?

2. If you do want it merged, I think some further unit tests for the Image.Equal and Images.Contains funcs wouldn't hurt. The updated e2e test gives some useful coverage but probably doesn't cover all cases

### DIFF
--- a/pkg/kbld/cmd/image.go
+++ b/pkg/kbld/cmd/image.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"reflect"
 	"strings"
 
 	"github.com/ghodss/yaml"
@@ -24,30 +25,6 @@ func (imgs Images) ForImage(url string) (Image, bool) {
 	return Image{}, false
 }
 
-func (i Image) Equal(other Image) bool {
-	if i.URL != other.URL {
-		return false
-	}
-	if len(i.Metas) != len(other.Metas) {
-		return false
-	}
-	for i, meta := range i.Metas {
-		if meta != other.Metas[i] {
-			return false
-		}
-	}
-	return true
-}
-
-func (imgs Images) Contains(img Image) bool {
-	for _, other := range imgs {
-		if img.Equal(other) {
-			return true
-		}
-	}
-	return false
-}
-
 // TODO only works after deserialization
 func (i Image) Description() string {
 	yamlBytes, err := yaml.Marshal(i.metasRaw)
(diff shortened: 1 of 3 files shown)


pkg/kbld/cmd/resolve.go
```diff
@@ -141,7 +141,14 @@ func (o *ResolveOptions) updateRefsInResources(nonConfigRs []ctlres.Resource,
...
 			if o.ImagesAnnotation {
-				images = append(images, img)
+				// if Metas is null then the image was already in digest form and we didn't need to resolve
+				// it, so the annotation isn't very useful
+				if len(img.Metas) > 0 {
+					// also check for duplicates before adding
+					if !Images(images).Contains(img) {
+						images = append(images, img)
+					}
+				}
```
cppforlife: seems like this is more specific to ResourceWithImages class. i think we should move deduping logic into its Bytes()

pkg/kbld/cmd/image.go
```diff
@@ -24,6 +24,30 @@ func (imgs Images) ForImage(url string) (Image, bool) {
 	return Image{}, false
 }
 
+func (i Image) Equal(other Image) bool {
+	if i.URL != other.URL {
```
cppforlife: i think we should move this into imageStruct object (https://github.com/k14s/kbld/blob/2c00de6b7df30b5214c797c1e91577d81b86031d/pkg/kbld/cmd/image.go#L37). comparison could be simplfied as well to something like this: `i.URL == other.URL && reflect.DeepEqual(i.Metas, other.Metas)`

jbrunton: Sounds good. Done.

pkg/kbld/cmd/resolve.go
jbrunton: I started doing this, but then I realized the imageStruct type is (unless I missed something) a supporting type used only by the ResourceWithImages class, so I wonder if it wouldn't make sense to keep the filtering in imageStruct instead (rather than have ResourceWithImages reach into the struct type)?

I've updated the draft in this way, but happy to move the filtering code into ResourceWithImages instead if you prefer – I'm still getting a sense of the relationships between classes.

(EDIT: In fact, if the imageStruct type is really a supporting type for ResourceWithImages, perhaps it would make sense to move that code to the ResourceWithImages file?)

PR #55, opened 2020-10-18
Feat/52 support additional image tags

### LINKED ISSUE #52: support uploading kbld built images to tags
For convenience, it may be helpful to allow the `destination.newImage` field of an `ImageDestination` document to contain a tag. Currently it errors when trying to auto-generate the upload tag

StarpTech:
Fix #52, the current implementation should serve as a first draft.

Configuration example:

```yaml
kind: ImageDestinations
destinations:
- image: demo-service
  newImage: docker.io/starptech/demo-service
  tags:
  - staging
  - v1.0.0
```

Related: https://github.com/k14s/kbld/issues/54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /kbld-windows-amd64.exe
 /tmp
 .DS_Store
+.vscode
--- a/pkg/kbld/config/config.go
+++ b/pkg/kbld/config/config.go
@@ -56,7 +56,8 @@ type ImageOverride struct {
 
 type ImageDestination struct {
 	ImageRef
-	NewImage string `json:"newImage"`
+	NewImage string   `json:"newImage"`
+	Tags     []string `json:"tags"`
 }
 
 type SearchRule struct {
--- a/pkg/kbld/config/config.go
+++ b/pkg/kbld/config/config.go
@@ -42,7 +42,6 @@ type Config struct {
 type Source struct {
 	ImageRef
 	Path string
-	Tags []string
 
 	Docker *SourceDockerOpts
 	Pack   *SourcePackOpts
--- a/pkg/kbld/config/config.go
+++ b/pkg/kbld/config/config.go
@@ -42,6 +42,7 @@ type Config struct {
 type Source struct {
 	ImageRef
 	Path string
+	Tags []string
 
 	Docker *SourceDockerOpts
 	Pack   *SourcePackOpts
--- a/pkg/kbld/image/built.go
+++ b/pkg/kbld/image/built.go
@@ -60,7 +60,6 @@ func (i BuiltImage) URL() (string, []ImageMeta, error) {
 			NoCache:    i.buildSource.Docker.Build.NoCache,
 			File:       i.buildSource.Docker.Build.File,
 			RawOptions: i.buildSource.Docker.Build.RawOptions,
-			Tags:       i.buildSource.Tags,
 		}
 
 		digest, err := i.docker.Build(urlRepo, i.buildSource.Path, opts)
--- a/pkg/kbld/image/built.go
+++ b/pkg/kbld/image/built.go
@@ -14,13 +14,12 @@ type BuiltImage struct {
 	buildSource ctlconf.Source
 	docker      Docker
 	pack        Pack
-	tags        []string
 }
 
 func NewBuiltImage(url string, buildSource ctlconf.Source,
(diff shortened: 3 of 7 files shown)


pivotal-issuemaster:
@StarpTech Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kbld&pullRequestId=55)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/kbld&pullRequestId=55) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@StarpTech Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kbld&pullRequestId=55)!

pivotaljohn:
Thanks for jumping on this, @StarpTech. 👍🏼 

Given that these tags are being applied to the image being created in the target registry, what are your thoughts of making the `tags` a child of `newImage`?

StarpTech:
> Given that these tags are being applied to the image being created in the target registry, what are your thoughts of making the tags a child of newImage?

Do you propose this?

```yaml
kind: ImageDestinations
destinations:
- image: demo-service
  newImage:
   image: docker.io/starptech/demo-service
   tags:
   - staging
   - v1.0.0
```

That's fine but I tried to avoid a breaking change.

cppforlife:
> That's fine but I tried to avoid a breaking change.

+1 to that.

also single array object is ImageDestination and to me tags make more sense as a root property.

cppforlife:
regarding general direction of this PR, i think we rely on a generic mechanism to tag (google/go-containerregistry library) as there is no guarantee that docker push will push under the same sha. on top of that since there will eventually be other builders that do not use docker we would want to have generic way of applying tags anyway.

StarpTech:
Any objections against this PR?

pivotaljohn:
Good reasons for keeping `tags` are an attribute of an ImageDestination; I'm clear on that point. 👍🏼

cppforlife:
> Any objections against this PR?

id like to see switching to google/go-containerregistry library (instead of using docker command) to add tags to pushed image.

StarpTech:
> id like to see switching to google/go-containerregistry library (instead of using docker command) to add tags to pushed image.

More specifically: You want to implement the standard https://github.com/opencontainers/distribution-spec/blob/master/spec.md? Is it already planned? As a contributor outside of the organization, it's hard to follow.

cppforlife:
> id like to see switching to google/go-containerregistry library (instead of using docker command) to add tags to pushed image.

to provide some context...

in docker builder, we use docker push to create random tag before we push because that the only way to identify what asset was pushed to the registry (https://github.com/k14s/kbld/blob/develop/pkg/kbld/image/docker.go#L168). that tagging behaviour is more of an internal impl (an unfortunate hack that we need).

since we want kbld to more with various builders (today we have docker and pack builders but we'll be adding builders that do not use docker at all) i would like to use [go-containerregistry library](https://github.com/google/go-containerregistry/) for applying user specified tags. this would allow us to be not coupled to docker for tagging functionality.

more concretely instead of using docker tag and docker push like the change you made in your PR i would expect to see code similar to this: https://github.com/google/go-containerregistry/blob/master/pkg/crane/tag.go to add tags. it probably should be in its own class `TaggedImage` and inserted here: https://github.com/k14s/kbld/blob/develop/pkg/kbld/image/factory.go#L44 like so `return NewTaggedImage(NewPushedImage(buildImg, imgDstConf, docker), imgDstConf)`

StarpTech:
I like the direction. Currently, my own use case affects docker only. It would be great to merge this feature. After this, we can investigate the broader approach. WDYT?

cppforlife:
> I like the direction. Currently, my own use case affects docker only. It would be great to merge this feature. After this, we can investigate the broader approach. WDYT?

ill give it a go tomorrow to see how complex switching to ggcr is.

cppforlife:
didnt turn out too complex: https://github.com/k14s/kbld/commit/79a1903ee8202c933c40edc61bdd5939bc8f1a6d#diff-cf6f5a33bdcae3ef8d18c07bf9726a2f98d77368103f7ff6f177edccb0137839. ill try to cut a new kbld release today.

PR #57, opened 2020-10-23
Feat/54 contr instr

StarpTech:


CONTRIBUTING.md
```diff
@@ -25,3 +25,55 @@ All contributors to this project must have a signed Contributor License Agreemen
...
+```
+
+## Run E2E tests against private docker registry
+```bash
+# Bootstrap k8s cluster
+minikube start
+# Build kbld binary for testing
+./hack/build.sh
+# Make your env aware of the docker registry
+eval $(minikube docker-env)
```
StarpTech: I'm not sure but this should not be required when using the docker registry? Instead, this should be followed https://minikube.sigs.k8s.io/docs/handbook/registry/#using-a-private-registry.

cppforlife: i typically use minikube _because_ it comes with docker daemon (since i dont install docker daemon directly on my machine). if you just have docker directly on your machine then none of the minikube stuff is needed since kbld does not know anything about k8s.

StarpTech: Good to know. Using docker hub is no good default because they have a fixed limit of 3 repositories for not paying customers (didn't work on my machine out of the box "ubuntu")

CONTRIBUTING.md
```diff
@@ -25,3 +25,41 @@ All contributors to this project must have a signed Contributor License Agreemen
...
+## Run Unit tests
+```bash
+# Run all tests
+./hack/test.sh
+# or run single test
+./hack/test.sh -run TestLogger
+```
+
+## Run E2E tests against minikube registry
+```bash
```
StarpTech, line 46: I think that's a good default.

cppforlife:
very appreciated for adding this information. ive moved these instructions into docs/dev.md (to be consistent with our other projects) and added https://github.com/k14s/kbld/blob/develop/docs/dev.md#source-code-structure.

StarpTech:
:+1:

PR #62, opened 2020-11-12
Feat/error check

StarpTech:
Uses https://github.com/kisielk/errcheck to lint for unhandled errors. I already add a list of excludes to not break the CI.


danielhelfand:
Curious if it would sense to add a more comprehensive linter than just errcheck? Maybe something like golangci-lint?

And then also eventually add it to the CI.

StarpTech:
I had that in mind but I'd like to introduce it incrementally. Err check is the most important one.

hack/lint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e -x -u
+
+GO111MODULE=off go get -u github.com/kisielk/errcheck
```
cppforlife: any reason why we are disabling go modules here? soon that will be a required mode afaik.

StarpTech: It should be installed globally not as a module in the workspace.

cppforlife: hmm, i guess this assumes that just by installing something with go it becomes available on the PATH. not sure if other folks have it set up that way, but in my case, above command does not make things show up on the path since my global $GOPATH/bin isnt part of PATH. if we are going to rely on GOPATH installed binary lets change line below to use something like `${GOPATH:-$(go env GOPATH)}/bin/errcheck`.

StarpTech: Adding `GOPATH/bin` to your PATH is part of the go installation [instructions](https://golang.org/doc/install#install). Also [`go install`](https://golang.org/cmd/go/#hdr-Compile_and_install_packages_and_dependencies) rely on that. I'd prefer to follow it it's a no-brainer.

StarpTech: But sure, we can do it :smile:

StarpTech:
> And then also eventually add it to the CI.

@danielhelfand it is part of the CI.

danielhelfand:
I think all the work added here would be really helpful, but it might make sense to have a discussion on possible linting tools/approaches before moving forward with this. I understand the incremental approach, but it could be interesting to bring this up as a community wide issue as I don't believe linting is in place across Carvel tools. It would be nice to have a consistent approach across all tools.

StarpTech:
> I think all the work added here would be really helpful, but it might make sense to have a discussion on possible linting tools/approaches before moving forward with this. I understand the incremental approach, but it could be interesting to bring this up as a community wide issue as I don't believe linting is in place across Carvel tools. It would be nice to have a consistent approach across all tools.

I'm open to a discussion but I'm not part of the internal team. Feel free to start a discussion. The reason why I added the errcheck is the way how this module is tested (E2E). An unhandled error could be swallowed and we won't know it until someone runs into an issue.

danielhelfand:
> I'm open to a discussion but I'm not part of the internal team. Feel free to start a discussion. The reason why I added the errcheck is the way how this module is tested (E2E). An unhandled error could be swallowed and we won't know it until someone runs into an issue.

Thanks for clarifying. My apologies for not seeing the intention. 

Perhaps it makes sense to move forward with this and use the conversation here to start the broader discussion. We can think of an appropriate space to discuss a long term linting approach.

StarpTech:
@danielhelfand :+1:

cppforlife:
(since this repo moved to vmware-tanzu org, we are blocked on setting up vmware CLA. will circle back on monday.)

StarpTech:
I found a much cleaner approach to deal with developer commands.
https://github.com/golang/go/issues/25922#issuecomment-402918061

cppforlife:
(@StarpTech already signed pivotal CLA; so no need to do vmware CLA for this PR)

PR #66, opened 2020-11-17
Update Vendor Files

### LINKED ISSUE #64: Is vendor directory used?
Hi, the vendor dir isn't in sync. Do we use `vendoring` in this project?

danielhelfand:
Co-Authored-By: Steven Locke <StevenLocke@users.noreply.github.com>

Closes #64 

As reported in the issue above, it appears vendor is out of sync. This pull request adds the files using `go mod vendor` to update the vendor folder.


StarpTech:
In order to benefit from vendoring we need to use `go build -mod vendor` in the scripts. What's the reason to do it at all? https://research.swtch.com/vgo-module#end_of_vendoring (by Russ Cox)

PR #73, opened 2020-11-25
Add builder creation prior to building images on buildkit

### LINKED ISSUE #69: Tests fail to execute in clean environment
When running the tests in a clean environment buildkit tests fail.

Reproduction:

This can be reproduced by hand.
1. Create a yaml file with:
```
kind: Object
spec:
- image: simple-app-two
- image: simple-app-three
---
kind: Sources
sources:
- image: simple-app-two
  path: assets/simple-app
  kubectlBuildkit:
    build:
      file: dev/Dockerfile.dev
- image: simple-app-three
  path: assets/simple-app
  kubectlBuildkit:
    build:
      target: build-env
```
2. Execute `cd test/e2e`
3. Run `../../kbld -f repro.yml --images-annotation=false`

The output we get is:
```
simple-app-three | starting build (using kubectl buildkit): assets/simple-app -> kbld:rand-1605907729199009000-163181625647-simple-app-three
simple-app-two | starting build (using kubectl buildkit): assets/simple-app -> kbld:rand-1605907729199007000-21210716319629-simple-app-two
simple-app-three | #1 [internal] booting buildkit
simple-app-two | #1 [internal] booting buildkit
simple-app-three | #1 ERROR: error while calling configMap.Create for "buildkit": configmaps "buildkit" already exists
simple-app-three | ------
simple-app-three |  > [internal] booting buildkit:
simple-app-three | ------
simple-app-three | Error: error while calling configMap.Create for "buildkit": configmaps "buildkit" already exists
simple-app-three | error: exit status 1
simple-app-three | finished build (using kubectl buildkit)
```

joaopapereira:
The `buildkit` tests are sometimes flaky due to the fact that we are
trying to create 2 images in parallel. When we try to do this `buildkit`
will try to create a builder, if it doesn't exist, in parallel which
causes a failure.

This commit ensures that the builder is created before we try to run the
parallel test to prevent the above-described issue

Closes #69

### DIFF
--- a/test/e2e/build_kubectl_buildkit_test.go
+++ b/test/e2e/build_kubectl_buildkit_test.go
@@ -5,6 +5,7 @@ package e2e
 
 import (
 	"fmt"
+	"os/exec"
 	"regexp"
 	"strings"
 	"testing"
@@ -13,6 +14,7 @@ import (
 func TestKubectlBuildkitBuildSuccessful(t *testing.T) {
 	env := BuildEnv(t)
 	kbld := Kbld{t, env.Namespace, env.KbldBinaryPath, Logger{}}
+	createBuilder(t)
 
 	input := env.WithRegistries(`
 kind: Object
@@ -61,6 +63,8 @@ func TestKubectlBuildkitBuildAndPushSuccessful(t *testing.T) {
 		return
 	}
 
+	createBuilder(t)
+
 	kbld := Kbld{t, env.Namespace, env.KbldBinaryPath, Logger{}}
 
 	input := env.WithRegistries(`
@@ -102,5 +106,21 @@ spec:
 
 	if out != expectedOut {
 		t.Fatalf("Expected >>>%s<<< to match >>>%s<<<", out, expectedOut)
+	}
+}
+
+// This function was created ensure that a builder was present in the cluster
+// before we try to run these tests.
+// If a builder does not exist the build command will try to create one.
+// The problem is that we run multiple build commands in parallel which cause
+// a race condition described in
+// https://github.com/vmware-tanzu/buildkit-cli-for-kubectl/issues/55
+// When this issue gets solved we should be able to remove this function
+func createBuilder(t *testing.T) {
+	cmd := exec.Command("kubectl", "buildkit", "create")
+
+	err := cmd.Run()
+	if err != nil {
+		t.Fatalf("error: %s\n", err)
 	}
 }


test/e2e/build_kubectl_buildkit_test.go
```diff
@@ -104,3 +109,16 @@ spec:
 		t.Fatalf("Expected >>>%s<<< to match >>>%s<<<", out, expectedOut)
 	}
 }
+
+func createBuilder(t *testing.T) {
```
danielhelfand, line 119: It might be helpful to create a `TODO` note here stating that this should be removed if/when it can be addressed upstream so it's not forgotten about.

danielhelfand: Resolving this comment. A separate issue will be opened regarding potentially fixing this upstream.

cppforlife: i would still prefer to add a comment to indicate why this exists and link it to github issue in buildkit repo.

test/e2e/build_kubectl_buildkit_test.go
```diff
@@ -104,3 +109,16 @@ spec:
...
+	var stdoutBuf, stderrBuf bytes.Buffer
+
+	cmd := exec.Command("kubectl", "buildkit", "create")
+	cmd.Stdout = &stdoutBuf
+	cmd.Stderr = &stderrBuf
+
+	err := cmd.Run()
+	if err != nil {
+		t.Fatalf(fmt.Sprintf("error: %s\n", err))
```
cppforlife: no need for fmt.Sprintf since we are using Fatalf

test/e2e/build_kubectl_buildkit_test.go
```diff
@@ -104,3 +109,16 @@ spec:
 		t.Fatalf("Expected >>>%s<<< to match >>>%s<<<", out, expectedOut)
 	}
 }
+
+func createBuilder(t *testing.T) {
+	var stdoutBuf, stderrBuf bytes.Buffer
+
+	cmd := exec.Command("kubectl", "buildkit", "create")
+	cmd.Stdout = &stdoutBuf
```
cppforlife: no need to set stdout or stderr.

test/e2e/build_kubectl_buildkit_test.go
joaopapereira: Added the comment

test/e2e/build_kubectl_buildkit_test.go
joaopapereira: fixed

test/e2e/build_kubectl_buildkit_test.go
joaopapereira: fixed

PR #98, opened 2021-02-03
Add ko support

### LINKED ISSUE #37: [builder] add ko as a backend
Some users want to use [`ko`](https://github.com/google/ko) as a backend (i.e. an image-building option).

Proposed Workflow:
- We want to preserve as much existing `kbld` behavior as possible, so we'll only use `ko` for building images and not for publishing images.
- To use `ko` for building, we can simply use `ko publish --local`. This will build images to the local Docker. `kbld` can then reference the local image for publishing.
- We expect this to integrate fairly easily in the existing code (similar to `pack`).

Example CR:
```
---
kind: Sources
sources:
- image: image1
  path: src/
  ko:
    build:
      rawOptions: ["--disable-optimizations"]
```

`ko.build.rawOptions ([]string)`: Refer to `ko publish -h` for all available flags

We didn't know what the rules/patterns were for docker and pack build options so we just opted for the simplest form.

benmoss:
Fixes #37 

With this yaml you can build kbld on this branch and then use kbld to make a kbld image using ko 😄 

```yaml
image: foo
---
kind: Sources
sources:
- image: foo
  path: ./cmd/kbld
  ko:
    build:
      rawOptions: ["--disable-optimizations"]
```

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -55,6 +55,9 @@ jobs:
         wget -O- https://github.com/buildpacks/pack/releases/download/v0.8.1/pack-v0.8.1-linux.tgz > /tmp/pack-cli.tgz
         tar xzvf /tmp/pack-cli.tgz -C /tmp/bin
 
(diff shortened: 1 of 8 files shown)


joaopapereira (approved):
LGTM

Issue #92, opened 2021-01-15
Deprecate functionality that is now in imgpkg


joaopapereira:
**Describe the problem/challenge you have**
Some of the functionality that already is part of `imgpkg` should be deprecated from `kbld`

**Describe the solution you'd like**
Scenario 1:

**Given** I have `kbld` installed
**When** I execute `kbld --help`
**Then** I should not see reference to `relocate`, `package` or `unpackage`

Scenario 2:

**Given** I have `kbld` installed
**When** I execute `kbld relocate --help`
**Then** I should see
```
Command "relocate" is deprecated, please use 'imgpkg copy', learn more in https://carvel.dev/imgpkg/docs/latest/commands/#copy
```

Scenario 3:

**Given** I have `kbld` installed
**When** I execute `kbld package --help`
**Then** I should see
```
Command "package" is deprecated, please use 'imgpkg push', learn more in https://carvel.dev/imgpkg/docs/latest/commands/#push
```

Scenario 4:

**Given** I have `kbld` installed
**When** I execute `kbld unpackage --help`
**Then** I should see
```
Command "unpackage" is deprecated, please use 'imgpkg pull', learn more in https://carvel.dev/imgpkg/docs/latest/commands/#pull
```

Scenario 5:

**Given** I access https://carvel.dev/kbld
**When** I click on the `Packaging & Relocation` link
**Then** I should see
```
Packaging and Relocation

This functionality was deprecated in `kbld` and will be removed in the near future. Check the new tool [imgpkg](https://carvel.dev/imgpkg).
```

**Anything else you would like to add:**
In this initial story, we should hide the commands and in the future remove them

cppforlife:
i wonder if it's also worth hiding pkg/unpkg/relocate commands from `kbld -h`? if not then may be also adding `[deprecated]` prefix to command description.

danielhelfand:
I am wondering if it would be worthwhile to write a message to stderr with a deprecation warning. Maybe also mentioning `imgpkg` docs in the message. This would be in addition to hiding the commands from help. 

Do we have any recommended time line for fully removing commands from other Carvel tools?

joaopapereira:
> I am wondering if it would be worthwhile to write a message to stderr with a deprecation warning. Maybe also mentioning imgpkg docs in the message. This would be in addition to hiding the commands from help.

Hmmm, I think we get that for free if we set the cobra command as deprecated, using the deprecated field in the struct. But it is good to ensure that.

[...]

PR #104, opened 2021-03-05
Mark package commands as deprecate

joaopapereira:
This marks the commands as deprecated.
I will create another PR for the documentation

Related to #92

### DIFF
--- a/pkg/kbld/cmd/package.go
+++ b/pkg/kbld/cmd/package.go
@@ -38,7 +38,13 @@ func NewPackageCmd(o *PackageOptions) *cobra.Command {
 		Use:     "package",
 		Aliases: []string{"pkg"},
 		Short:   "Package configuration and images into tarball",
-		RunE:    func(_ *cobra.Command, _ []string) error { return o.Run() },
+		Hidden:  true,
+		Long: `
+Command "package" is deprecated, please use 'imgpkg push', learn more in https://carvel.dev/imgpkg/docs/latest/commands/#push
+
+Package configuration and images into tarball
+`,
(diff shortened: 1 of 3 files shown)


pkg/kbld/cmd/package.go
```diff
@@ -48,6 +53,12 @@ func NewPackageCmd(o *PackageOptions) *cobra.Command {
 }
 
 func (o *PackageOptions) Run() error {
+	_, err := fmt.Fprintln(os.Stderr, `Command "package" is deprecated, please use 'imgpkg push', learn more in https://carvel.dev/imgpkg/docs/latest/commands/#push`)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintln(os.Stderr, `NOOOOOOOOOOOOOOOO`)
```
cppforlife: dont really want this.

pkg/kbld/cmd/package.go
```diff
@@ -48,6 +53,12 @@ func NewPackageCmd(o *PackageOptions) *cobra.Command {
 }
 
 func (o *PackageOptions) Run() error {
+	_, err := fmt.Fprintln(os.Stderr, `Command "package" is deprecated, please use 'imgpkg push', learn more in https://carvel.dev/imgpkg/docs/latest/commands/#push`)
```
cppforlife: we got `logger := ctllog.NewLogger(os.Stderr)` a few lines below. let's use that to print this message. let's also prefix it with `Warning:`

pkg/kbld/cmd/package.go
joaopapereira: lol, this was just me testing to ensure the correct commit was being tested :p

Issue #109, opened 2021-05-13
Manifest searching stops recursing below any key named `image`


coro:
**What steps did you take:**
To demonstrate the issue, here is a repro, containing a kbld config, an imgpkg lockfile and a manifest I wish to transform:
```yaml
---
apiVersion: kbld.k14s.io/v1alpha1
kind: Config
minimumRequiredVersion: 0.29.0
searchRules:
- keyMatcher:
    path: [spec, image, default]
- keyMatcher:
    path: [spec, nonimage, default]
---
apiVersion: imgpkg.carvel.dev/v1alpha1
images:
- annotations:
    kbld.carvel.dev/id: original-image:1.2.3
  image: updated-image@sha256:2005721c33a949d82bab23099aa3f577932c350bc24d5931932e7440856f6276
kind: ImagesLock
---
spec:
  image:
    default: original-image:1.2.3
  nonimage:
    default: original-image:1.2.3
```

**What happened:**
I ran the kbld command, and only some of the image references were replaced with the resolved image path:
```bash
$ kbld -f manifest.yaml
resolve | final: original-image:1.2.3 -> updated-image@sha256:2005721c33a949d82bab23099aa3f577932c350bc24d5931932e7440856f6276
---
metadata:
  annotations:
    kbld.k14s.io/images: |
      - Metas:
        - Type: preresolved
          URL: updated-image@sha256:2005721c33a949d82bab23099aa3f577932c350bc24d5931932e7440856f6276
        URL: updated-image@sha256:2005721c33a949d82bab23099aa3f577932c350bc24d5931932e7440856f6276
spec:
  image:
    default: original-image:1.2.3 # This is left not updated
  nonimage:
    default: updated-image@sha256:2005721c33a949d82bab23099aa3f577932c350bc24d5931932e7440856f6276 # This is updated
```

This also occurs if you use a valueMatcher for all instances of the image:
```yaml
---
apiVersion: kbld.k14s.io/v1alpha1
kind: Config
minimumRequiredVersion: 0.29.0
searchRules:
- valueMatcher:
    image: original-image:1.2.3
---
apiVersion: imgpkg.carvel.dev/v1alpha1
images:
- annotations:
    kbld.carvel.dev/id: original-image:1.2.3
  image: updated-image@sha256:2005721c33a949d82bab23099aa3f577932c350bc24d5931932e7440856f6276
kind: ImagesLock
---
spec:
  image:
    default: original-image:1.2.3
  nonimage:
[...]

PR #110, opened 2021-05-14
if update strategy cannot handle matched type, continue with searching

cppforlife:
solves https://github.com/vmware-tanzu/carvel-kbld/issues/109


joaopapereira (approved):
LGTM

PR #111, opened 2021-06-03
Update kbld to go 1.16

danielhelfand:
Updating to standardize version of go used across Carvel tools


joaopapereira (changes requested):
The Github actions also have the Go version. We should change them as well

danielhelfand:
> The Github actions also have the Go version. We should change them as well

Thanks, have gone ahead and updated

PR #130, opened 2021-07-19
Update readme to include kubectl-buildkit dependency

cari-lynn:
kubectl-buildkit is now a prerequisite to running the e2e tests
https://github.com/vmware-tanzu/carvel-kbld/blob/develop/hack/test-e2e.sh#L10-L13


cari-lynn:
I see the readme has instructions to run the tests with minicube, that should take care of these prereqs

PR #132, opened 2021-07-22
Add metadata in imgpkg lock file

cari-lynn:
**When using the `--imgpkg-lock-output` flag, this PR adds the ` kbld.carvel.dev/metas` annotation.**
```
images:
- annotations:
    kbld.carvel.dev/metas: |
      - Tag: x.x.x
        Type: resolved
        URL: image-name:x.x.x
```

**When using a lock file as input to kbld, this PR also will preserve any ` kbld.carvel.dev/metas`**. It will append new metas to the list in the output and an imgpkg Lock file. This results in a history for example:
```
images:
- annotations:
    kbld.carvel.dev/metas: |
      - Tag: x.x.x
        Type: resolved
        URL: image-name:x.x.x
      - Type: preresolved
        URL: index.docker.io/library/..@sha256:...
```

**Other new behavior**:
- If a ` kbld.carvel.dev/metas` annotation is malformatted, kbld will error.
- the ` kbld.carvel.dev/metas` annotation must be in the format of several predefined objects, other unknown meta objects will be ignored.

### DIFF
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -235,8 +235,6 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 			})
 		}
 
-		c.Overrides = ctlconf.UniqueImageOverrides(c.Overrides)
-
 		return c.WriteToFile(o.LockOutput)
 	case o.ImgpkgLockOutput != "":
 		iLock := lockconfig.ImagesLock{
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -246,7 +246,7 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 			},
 		}
 		for _, urlImagePair := range resolvedImages.All() {
-			anns := imgpkgLockAnnotations(urlImagePair)
+			anns := o.imgpkgLockAnnotations(urlImagePair)
 
 			iLock.Images = append(iLock.Images, lockconfig.ImageRef{
 				Image:       urlImagePair.Image.URL,
@@ -259,7 +259,7 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 	}
 }
 
-func imgpkgLockAnnotations(i ProcessedImageItem) map[string]string {
+func (o *ResolveOptions) imgpkgLockAnnotations(i ProcessedImageItem) map[string]string {
 	anns := map[string]string{
 		ctlconf.ImagesLockKbldID: i.UnprocessedImageURL.URL,
 	}
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -263,14 +263,9 @@ func (o *ResolveOptions) imgpkgLockAnnotations(i ProcessedImageItem) map[string]
 	anns := map[string]string{
 		ctlconf.ImagesLockKbldID: i.UnprocessedImageURL.URL,
 	}
-
-	imageMetas := i.Image.MetasDescription()
-	if imageMetas != nil && len(imageMetas) > 0 {
-		var metas []ctlconf.ImageMeta
-		for _, m := range imageMetas {
-			metas = append(metas, m)
-		}
-		metaYaml, err := yaml.Marshal(metas)
+	imageMetas := i.Metas
+	if len(imageMetas) > 0 {
+		metaYaml, err := yaml.Marshal(imageMetas)
 		if err != nil {
 			return anns
 		}
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -266,13 +266,15 @@ func (o *ResolveOptions) imgpkgLockAnnotations(i ProcessedImageItem) map[string]
 
 	imageMetas := i.Image.MetasDescription()
 	if imageMetas != nil && len(imageMetas) > 0 {
-		metaYaml, err := yaml.Marshal([]ctlconf.ImageMeta{
-			{
-				URL:  imageMetas[0].URL,
-				Type: "preresolved",
-				Tag:  imageMetas[0].Tag,
-			},
-		})
+		var metas []ctlconf.ImageMeta
+		for _, m := range imageMetas {
+			metas = append(metas, ctlconf.ImageMeta{
+				URL:  m.URL,
+				Type: m.Type,
+				Tag:  m.Tag,
+			})
+		}
+		metaYaml, err := yaml.Marshal(metas)
 		if err != nil {
 			return anns
 		}
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -20,6 +20,7 @@ import (
 	ctlser "github.com/k14s/kbld/pkg/kbld/search"
 	"github.com/k14s/kbld/pkg/kbld/version"
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
 )
 
 type ResolveOptions struct {
@@ -245,9 +246,16 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 			},
 		}
 		for _, urlImagePair := range resolvedImages.All() {
+			imagesYAML, err := yaml.Marshal(urlImagePair.Image.Metas)
+			if err != nil {
+				return err
+			}
 			iLock.Images = append(iLock.Images, lockconfig.ImageRef{
-				Image:       urlImagePair.Image.URL,
-				Annotations: map[string]string{ctlconf.ImagesLockKbldID: urlImagePair.UnprocessedImageURL.URL},
+				Image: urlImagePair.Image.URL,
+				Annotations: map[string]string{
+					ctlconf.ImagesLockKbldID:    urlImagePair.UnprocessedImageURL.URL,
+					ctlconf.ImagesLockKbldMetas: string(imagesYAML),
+				},
 			})
 		}
 		return iLock.WriteToPath(o.ImgpkgLockOutput)
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -246,20 +246,38 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 			},
 		}
 		for _, urlImagePair := range resolvedImages.All() {
-			imagesYAML, err := yaml.Marshal(urlImagePair.Image.Metas)
-			if err != nil {
-				return err
-			}
+			anns := imgpkgLockAnnotations(urlImagePair)
+
 			iLock.Images = append(iLock.Images, lockconfig.ImageRef{
-				Image: urlImagePair.Image.URL,
-				Annotations: map[string]string{
-					ctlconf.ImagesLockKbldID:    urlImagePair.UnprocessedImageURL.URL,
-					ctlconf.ImagesLockKbldMetas: string(imagesYAML),
-				},
+				Image:       urlImagePair.Image.URL,
+				Annotations: anns,
 			})
 		}
 		return iLock.WriteToPath(o.ImgpkgLockOutput)
 	default:
 		return nil
 	}
+}
+
+func imgpkgLockAnnotations(i ProcessedImageItem) map[string]string {
+	anns := map[string]string{
+		ctlconf.ImagesLockKbldID: i.UnprocessedImageURL.URL,
+	}
+
+	imageMetas := i.Image.MetasDescription()
+	if imageMetas != nil && len(imageMetas) > 0 {
+		metaYaml, err := yaml.Marshal([]ctlconf.ImageMeta{
+			{
+				URL:  imageMetas[0].URL,
+				Type: "preresolved",
+				Tag:  imageMetas[0].Tag,
+			},
+		})
+		if err != nil {
+			return anns
+		}
+		anns[ctlconf.ImagesLockKbldMetas] = string(metaYaml)
+	}
+
+	return anns
 }
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -268,11 +268,7 @@ func (o *ResolveOptions) imgpkgLockAnnotations(i ProcessedImageItem) map[string]
 	if imageMetas != nil && len(imageMetas) > 0 {
 		var metas []ctlconf.ImageMeta
 		for _, m := range imageMetas {
-			metas = append(metas, ctlconf.ImageMeta{
-				URL:  m.URL,
-				Type: m.Type,
-				Tag:  m.Tag,
-			})
+			metas = append(metas, m)
 		}
 		metaYaml, err := yaml.Marshal(metas)
 		if err != nil {
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -246,11 +246,9 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 			},
 		}
 		for _, urlImagePair := range resolvedImages.All() {
-			anns := o.imgpkgLockAnnotations(urlImagePair)
-
 			iLock.Images = append(iLock.Images, lockconfig.ImageRef{
 				Image:       urlImagePair.Image.URL,
-				Annotations: anns,
+				Annotations: o.imgpkgLockAnnotations(urlImagePair),
 			})
 		}
 		return iLock.WriteToPath(o.ImgpkgLockOutput)
@@ -263,13 +261,12 @@ func (o *ResolveOptions) imgpkgLockAnnotations(i ProcessedImageItem) map[string]
 	anns := map[string]string{
 		ctlconf.ImagesLockKbldID: i.UnprocessedImageURL.URL,
 	}
-	imageMetas := i.Metas
-	if len(imageMetas) > 0 {
-		metaYaml, err := yaml.Marshal(imageMetas)
+	if len(i.Metas) > 0 {
+		bs, err := yaml.Marshal(i.Metas)
 		if err != nil {
 			return anns
 		}
-		anns[ctlconf.ImagesLockKbldMetas] = string(metaYaml)
+		anns[ctlconf.ImagesLockKbldMetas] = string(bs)
 	}
 
 	return anns
--- a/pkg/kbld/config/config.go
+++ b/pkg/kbld/config/config.go
@@ -73,9 +73,14 @@ type ImageOverride struct {
 }
 
 type ImageMeta struct {
-	URL  string `json:"URL,omitempty" yaml:"URL,omitempty"`
-	Type string `json:"Type,omitempty" yaml:"Type,omitempty"`
-	Tag  string `json:"Tag,omitempty" yaml:"Tag,omitempty"`
+	URL       string   `json:",omitempty" yaml:",omitempty"`
+	Type      string   `json:",omitempty" yaml:",omitempty"`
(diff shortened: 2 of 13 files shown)


gcheadle-vmware:
@cari-lynn Requesting review, was unable to select you as a GH reviewer.

pkg/kbld/cmd/resolve.go
```diff
@@ -245,13 +246,38 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
...
+				Annotations: anns,
 			})
 		}
 		return iLock.WriteToPath(o.ImgpkgLockOutput)
 	default:
 		return nil
 	}
 }
+
+func imgpkgLockAnnotations(i ProcessedImageItem) map[string]string {
```
cppforlife: lets place that on ResolveOptions so that it's not package global.

pkg/kbld/config/config.go
```diff
@@ -67,6 +68,13 @@ type ImageOverride struct {
 	NewImage     string                     `json:"newImage"`
 	Preresolved  bool                       `json:"preresolved,omitempty"`
 	TagSelection *versions.VersionSelection `json:"tagSelection,omitempty"`
+	ImageMetas   []ImageMeta                `json:",omitempty"`
+}
+
+type ImageMeta struct {
+	URL  string `json:"URL,omitempty" yaml:"URL,omitempty"`
+	Type string `json:"Type,omitempty" yaml:"Type,omitempty"`
+	Tag  string `json:"Tag,omitempty" yaml:"Tag,omitempty"`
```
cppforlife: i was under impression that some metas have other fields. i think we need to make this more generic.

pkg/kbld/config/config.go
```diff
@@ -167,12 +175,15 @@ func NewConfigFromImagesLock(res ctlres.Resource) (Config, error) {
 	overridesConfig := NewConfig()
 
 	for _, image := range imagesLock.Images {
+		var iMetas []ImageMeta
+		yaml.Unmarshal([]byte(image.Annotations[ImagesLockKbldMetas]), &iMetas)
```
cppforlife: yaml.Unmarshal errors. what should we do with it?

pkg/kbld/image/preresolved.go
```diff
@@ -3,21 +3,34 @@
...
+func NewPreresolvedImage(url string, metas []config.ImageMeta) PreresolvedImage {
+	return PreresolvedImage{url, metas}
 }
 
 func (i PreresolvedImage) URL() (string, []Meta, error) {
+	if len(i.metas) > 0 {
+		var givenMetas []Meta
+		for _, m := range i.metas {
+			givenMetas = append(givenMetas, PreresolvedImageSourceURL{Type: "preresolved", URL: m.URL, Tag: m.Tag})
+		}
```
cppforlife: not sure why we would be setting type to preresolved on given metas. if there are metas that indicate that it's git, it should remain as is. additionally we should append preresolved meta from line 35 to the results.

test/e2e/lock_output_test.go
```diff
@@ -208,3 +227,78 @@ images:
...
+      - Metas:
+        - Tag: 1.14.2
+          Type: preresolved
+          URL: nginx:1.14.2
+        URL: index.docker.io/library/nginx@sha256:f7988fb6c02e0ce69257d9bd9cf37ae20a60f1df7563c3a2a6abe24160306b8d
+`
+	if out != expectedOut {
+		t.Fatalf("Expected >>>%s<<< to match >>>%s<<<", out, expectedOut)
+	}
+
```
cppforlife, line 350: nit: remove newline

pkg/kbld/config/config.go
cari-lynn: If an imgpkg lock file cannot be marshalled into an ImageMeta struct, we should report an error that the lock file format was incorrect.

pkg/kbld/image/preresolved.go
```diff
@@ -3,21 +3,34 @@
...
+		return i.url, givenMetas, nil
+	}
```
cari-lynn: Rename givenMetas -> imageMetas? Since 'given' doesn't map to anything related to this domain.

pkg/kbld/image/preresolved.go
```diff
@@ -3,21 +3,21 @@
...
+	return PreresolvedImage{url, metas}
 }
 
-func (PreresolvedImageSourceURL) meta() {}
-
-func NewPreresolvedImage(url string) PreresolvedImage {
-	return PreresolvedImage{url}
-}
+func (i PreresolvedImage) URL() (string, []ctlconf.Meta, error) {
+	imageMetas := append(i.metas, ctlconf.PreresolvedImageSourceURL{Type: "preresolved", URL: i.url})
```
cppforlife: why do we need to mutate PreresolvedImage?

pkg/kbld/image/preresolved.go
```diff
@@ -3,21 +3,21 @@
...
-	url string
+	url   string
+	metas []ctlconf.Meta
 }
 
-type PreresolvedImageSourceURL struct {
-	Type string // always set to 'preresolved'
-	URL  string
+func NewPreresolvedImage(url string, metas []ctlconf.Meta) PreresolvedImage {
+	return PreresolvedImage{url, metas}
```
cppforlife: we should make a copy of metas here to avoid accidentally mutating given array.

pkg/kbld/config/image_meta.go
```diff
@@ -0,0 +1,97 @@
...
+			m.Metas = append(m.Metas, local)
+		case yaml.Unmarshal(yamlItem, &git) == nil && git.Type == "git":
+			m.Metas = append(m.Metas, git)
+		case yaml.Unmarshal(yamlItem, &res) == nil && res.Type == "resolved":
+			m.Metas = append(m.Metas, res)
+		case yaml.Unmarshal(yamlItem, &preres) == nil && preres.Type == "preresolved":
+			m.Metas = append(m.Metas, preres)
+		case yaml.Unmarshal(yamlItem, &tag) == nil && tag.Type == "tagged":
+			m.Metas = append(m.Metas, tag)
+		}
```
cppforlife, line 117: lets add 

```
default:
  // ignore unknown type
```

pkg/kbld/config/image_meta.go
```diff
@@ -0,0 +1,97 @@
...
+		var local BuiltImageSourceLocal
+		var git BuiltImageSourceGit
+		var res ResolvedImageSourceURL
+		var preres PreresolvedImageSourceURL
+		var tag TaggedImageMeta
+
+		yamlItem, _ := yaml.Marshal(&item)
+
+		switch {
+		case yaml.Unmarshal(yamlItem, &local) == nil && local.Type == "local":
```
cppforlife: should probably set these as consts since they are used elsewhere. (though ideally one wouldnt have to set Type manually at all)

pkg/kbld/config/image_meta.go
```diff
@@ -0,0 +1,97 @@
...
+	"encoding/json"
+
+	"sigs.k8s.io/yaml"
+)
+
+type Meta interface {
+	meta()
+}
+
+type ImageMeta struct {
```
cppforlife: it seems that this type is used purely for unmarshaling. we should make it private to this package then.

pkg/kbld/config/image_meta.go
```diff
@@ -0,0 +1,97 @@
...
+	Tag  string `json:",omitempty" yaml:",omitempty"`
+}
+
+func (BuiltImageSourceGit) meta()       {}
+func (BuiltImageSourceLocal) meta()     {}
+func (ResolvedImageSourceURL) meta()    {}
+func (TaggedImageMeta) meta()           {}
+func (PreresolvedImageSourceURL) meta() {}
+
+func metasHistory(metas string) ([]Meta, error) {
```
cppforlife: metasHistory -> NewMetasFromString

pkg/kbld/config/config.go
```diff
@@ -322,7 +329,8 @@ func UniqueImageOverrides(overrides []ImageOverride) []ImageOverride {
 	for _, override := range overrides {
 		var found bool
 		for _, addedOverride := range result {
-			if addedOverride == override {
+			// using DeepEqual instead of '==' since ImageOverride contains a slice
+			if reflect.DeepEqual(addedOverride, override) {
```
cppforlife: do we have any funky behaviour when we have same image multime times with different metas?

pkg/kbld/config/config.go
```diff
@@ -67,6 +68,7 @@ type ImageOverride struct {
 	NewImage     string                     `json:"newImage"`
 	Preresolved  bool                       `json:"preresolved,omitempty"`
 	TagSelection *versions.VersionSelection `json:"tagSelection,omitempty"`
+	ImageMetas   []Meta                     `json:",omitempty"`
```
cppforlife: seems like we should be specifying serialized name explicitly here. i do wonder if we should call it just metas though.

pkg/kbld/cmd/resolve.go
```diff
@@ -245,13 +246,31 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
...
 	default:
 		return nil
 	}
 }
+
+func (o *ResolveOptions) imgpkgLockAnnotations(i ProcessedImageItem) map[string]string {
+	anns := map[string]string{
+		ctlconf.ImagesLockKbldID: i.UnprocessedImageURL.URL,
+	}
+	imageMetas := i.Metas
```
cppforlife: not sure why we need local variable. use i.Metas directly?

pkg/kbld/cmd/resolve.go
```diff
@@ -245,13 +246,31 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
...
+	if len(imageMetas) > 0 {
+		metaYaml, err := yaml.Marshal(imageMetas)
```
cppforlife: metaYaml -> metasYAML (or my favorite `bs`)

gcheadle-vmware:
This looks good to me. Great idea extracting out the meta interface into its own package, it really cleaned up nicely :)

pkg/kbld/cmd/resolve.go
```diff
@@ -245,13 +246,31 @@ func (o *ResolveOptions) emitLockOutput(conf ctlconf.Conf, resolvedImages *Proce
 			},
 		}
 		for _, urlImagePair := range resolvedImages.All() {
+			anns := o.imgpkgLockAnnotations(urlImagePair)
+
 			iLock.Images = append(iLock.Images, lockconfig.ImageRef{
 				Image:       urlImagePair.Image.URL,
-				Annotations: map[string]string{ctlconf.ImagesLockKbldID: urlImagePair.UnprocessedImageURL.URL},
+				Annotations: anns,
```
pivotaljohn: **nit:** inline `anns`.

**rationale:**
- the variable `anns` doesn't seem to be doing much work (it's not reducing repeated invocations or saving memory, nor is it improving the readability)
- the function does not return an `error`
- the function describes well the returned result

pkg/kbld/image/resolved.go
```diff
@@ -7,28 +7,21 @@ import (
 	"fmt"
 
 	regname "github.com/google/go-containerregistry/pkg/name"
+	ctlconf "github.com/k14s/kbld/pkg/kbld/config"
 	ctlreg "github.com/k14s/kbld/pkg/kbld/registry"
 )
 
-// ResolvedImage respresents an image that will be resolved into url+digest
+// ResolvedImage represents an image that will be resolved into url+digest
```
pivotaljohn, line 14: ❤️

pkg/kbld/image/tag_selected.go
```diff
@@ -7,11 +7,12 @@ import (
 	"fmt"
 
 	regname "github.com/google/go-containerregistry/pkg/name"
+	ctlconf "github.com/k14s/kbld/pkg/kbld/config"
 	ctlreg "github.com/k14s/kbld/pkg/kbld/registry"
 	versions "github.com/vmware-tanzu/carvel-vendir/pkg/vendir/versions/v1alpha1"
 )
 
-// TagSelectedImage respresents an image that will be resolved into url+digest
+// TagSelectedImage represents an image that will be resolved into url+digest
```
pivotaljohn, line 15: 😆

pkg/kbld/image/tagged.go
```diff
@@ -9,25 +9,18 @@ import (
 	ctlreg "github.com/k14s/kbld/pkg/kbld/registry"
 )
 
-// TaggedImage respresents an image that will be tagged when its URL is requested
+// TaggedImage represents an image that will be tagged when its URL is requested
```
pivotaljohn, line 12: clearly copy pasta. Thanks for pulling these weeds.

pkg/kbld/config/image_meta.go
pivotaljohn:
```suggestion
func parseMetas(metas string) ([]Meta, error) {
```

It would be great if "Meta" had a better metaphor... and what I'm seeing here is an attempt to suggest one: something less reliable than "provenance" but tells a story from whence this image came: a history.

However, in _this_ context, this function is effectively taking the unmarshalling just one more step to include these `Meta` types. So, the intention of the function actually gets a bit blurred.

Another angle to "justify" this suggestion is that functions often _tend to_ work better as verb[-phrases] than nouns.

"Parse" rather than "unmarshall" because the function accepts an unstructured input and _produces_ a structured output, rather than stuffing a structure. 🤷🏻

pkg/kbld/config/image_meta.go
pivotaljohn: Should we be deliberately ignoring unmarshallable meta?

I ask because when we ignore these errors some piece of this "history" is dropped without warning.
And while Carvel tools do not make direct use of this data, it is in effect an API; other tooling may well become dependent on the presence of it.

An alternative is to error out. If the user wants to ignore such an error, they can remove the meta annotations and try again.

pivotaljohn (changes requested):
This work is _really_ close.

The overall approach looks really good to me.

Details in the comments, below.

pivotaljohn:
Reviewed w/ @garrettcheadle (offering to make changes if desired):

> i think that all suggestions but the last comment are simple changes that I see as improvements. ... If you could make those changes/renames, that would move this along.

👍🏻 

> In the case of "[deliberately ignoring unmarshallable data](https://github.com/vmware-tanzu/carvel-kbld/pull/132#discussion_r707772960)" ... we have another custom unmarshaller in kbld, maybe we should check if that unmarshaller has an error case for wisdom?

I'll look into that.

pivotaljohn:
In terms of prior art, I'm seeing this idiom which attempts to unmarshal data into a set of potential structs...

https://github.com/vmware-tanzu/carvel-kbld/blob/f919f7f78a/pkg/kbld/resources/path.go#L78-L91

By the end of the `switch`, if no suitable struct was found, that was reported as an error.

I think we should do a similar thing in the case of detecting which kind of `Meta` (if you can successfully unmarshall it, must be _that_ kind!)... but also either emit a warning or error out when it's unrecognized.

cppforlife (commented):
(looks like i had review that wasnt submitted. might be a bit stale.)

pkg/kbld/config/config.go
pivotaljohn: Yeah, we want to look into what the possible "conflict" use cases there are... and how we want each to be resolved.  And if need be, make the appropriate changes.

pkg/kbld/config/config.go
pivotaljohn: Yeah, avoids accidental "breaking changes" if we rename the struct field.

I wish we had a more descriptive name than "metas:"

pkg/kbld/config/image_meta.go
pivotaljohn: See https://github.com/vmware-tanzu/carvel-kbld/pull/132#discussion_r707772960 for resolution.

pkg/kbld/config/image_meta.go
pivotaljohn: For now, dropping image meta (which plays the role of "resolution history"), is the way to go:
- meta are diagnostic information, not operational data
- reporting errors well (e.g. not flooding the output with warnings) adds complexity that's not worth the trade-off right now
- captured the possibility of doing more here: #160.

Added a comment noting that we're deliberately dropping meta (and why).

pkg/kbld/image/preresolved.go
pivotaljohn: Making a copy before appending. 👍🏻

pkg/kbld/config/image_meta.go
pivotaljohn: done.

pkg/kbld/config/image_meta.go
pivotaljohn: done.

pkg/kbld/config/image_meta.go
pivotaljohn: done.

pkg/kbld/config/config.go
pivotaljohn: added "metas" for now; will potentially rename in a separate commit.

pkg/kbld/cmd/resolve.go
pivotaljohn: done.

pkg/kbld/cmd/resolve.go
pivotaljohn: that's just bullsh*t!

... done.

pkg/kbld/config/config.go
pivotaljohn: ~~I think it looks like this:~~

- ~~no dups ==> no overrides removed~~
- ~~matching overrides (exact duplicates) ==> duplicates removed~~
- ~~equal "image[+imagerepo]":~~
  - ~~preresolved > !preresolved — respect presence of "preresolved" by not re-resolving~~
  - ~~!tagselection > tagselection — presume "newImage" without tagselection is more precise and therefore wins~~
  - ~~latest(tagselection) — for `semver` (current only strategy supported), whichever selects the largest/latest version value.~~
- ~~first override specified wins — if a set of overrides are not unique by this point, then take the first.~~

---
_edit: I misunderstood the intent, here; revised in next comment._

---
_edit: in fact, this is a bad idea, overall: it removes the control from consuming users who might want to truly override._

pivotaljohn: oh wait... I think I misunderstood our intent, here.

- on reflection, I see that: _The order_ of overrides is the only thing that determines which one applies (i.e. first _always_ wins, regardless) — this is **not** changing
- ... rather ...: There are a couple of cases where additional meta are added (notably "preresolve"s). The concern here is making sure that we don't lose any meta on the way through de-duping. 
- also noted while reading the surrounding code more carefully: we don't actually call `UniqueImageOverrides()` when emitting an `imgpkg`-flavored lock file. I believe this is an oversight.

So, I'll review this flow with my revised understanding, today.

pivotaljohn: Back to the original question:
> do we have any funky behaviour when we have same image multime times with different metas?

The answer is no... mostly because in all the code paths that lead to `UniqueImageOverrides()`, _right now_, none of the overrides have meta set.

This led me to step back and reflect on the overall workflow...

With this PR, we are propagating meta through `imgpkg` lock files.
However, we are _not_ doing so with plain `kbld` lock files.
Technically, this was not requested in either the issue or intended in this PR.

That said, if it's not unreasonable that users may which to use `kbld` lock files during development and then switch to `imgpkg` lock files when bundling up the config+images, then it would make sense to ensure that meta are carried through the `kbld` lock file flow.

I'm capturing this additional scope in a separate issue and it can be prioritized: #161.

pivotaljohn (approved):

pivotaljohn:
Reviewed w/ @cppforlife over the phone.

- [x] `UniqueImageOverrides` is not needed when preparing to output `imgpkg`-flavored lock files. The other callers are in deprecated code. We'll remove this invocation from "resolve"
- [ ] to minimize the impact of this change on working ocde, we'll fix the compilation error, by writing a custom `Equals()` that ignores metas in the comparison. 👍🏻

Issue #128, opened 2021-07-13
kbld integration with bazel for building imgpkg bundles. 


mohansitaram:
**Describe the problem/challenge you have**
We are currently using bazel to build docker images from source code. We are also using helm charts and imgpkg bundles to package all application code and config. The challenge we are facing is this: Given an application that has a helm chart and bazel rules for building its docker images, how do we create an imgpkg bundle for it?

**Describe the solution you'd like**

There are a couple of ways to do this:

1. No direct integration between kbld and bazel:  
- Build image using bazel directly and push this image to an intermediate registry. 
- Update helm chart to use the above image. Then build imgpkg bundle for the application.
2. With kbld and bazel integration: 
- kbld calls bazel and builds the docker image. 
- kbld rewrites the image reference in the helm chart to point to the docker image on the local docker engine (i.e. the image that gets built using bazel in previous step). Then build imgpkg bundle for the application. 

**Anything else you would like to add:**
We could write custom tooling around kbld and bazel to implement solution 1 but solution 2 would be ideal. That is a solution in which kbld provides config or hooks wherein we can specify the location of the bazel config to use for building images.

We would also like to avoid the push of the docker images to the intermediate registry. The docker images would be already built using bazel and available on the local docker engine so it would be great if there is a way to use kbld to overwrite image references in helm charts to point to the local images. I believe this is already possible with other tools like pack and buildkit but I have not used those tools so I could be wrong. 

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

aaronshurley:
Hi @mohansitaram, thank you for surfacing this idea. I can see how the streamlined workflow in the second scenario that you proposed would be an improvement. Personally, I'm not very familiar with bazel but if I understand your use then it feels similar to the way that kbld currently [allows different build sources (such as docker, pack, buildkit, ko)](https://carvel.dev/kbld/docs/latest/config/#sources). Would you expect that kbld would only need to run any commands other than `bazel build`?

> We would also like to avoid the push of the docker images to the intermediate registry.

[...]

PR #148, opened 2021-08-28
Add Bazel support

cliuj:
Implements option 2 stated in: https://github.com/vmware-tanzu/carvel-kbld/issues/128.

The implementation is similar to the other builders. `bazel run <label>` is used for the sub process command as it outputs the image digest if `container_image(...)` is used. 

One thing to note about `bazel run <label>` is that it isn't limited to just image building, but can also run tests and binaries (depending on how its configured). Thus it is recommended that the label used for `label` has `container_image(...)` as the last step in the `BUILD.bazel`.


cliuj:
Closing this PR to update CLA status.

PR #149, opened 2021-08-28
Add Bazel support

cliuj:
(Re-opened due to CLA issue)

Implements option 2 stated in: https://github.com/vmware-tanzu/carvel-kbld/issues/128.

The implementation is similar to the other builders. `bazel run <label>` is used for the sub process command as it outputs the image digest if `container_image(...)` is used. 

One thing to note about `bazel run <label>` is that it isn't limited to just image building, but can also run tests and binaries (depending on how its configured). Thus it is recommended that the label used for `label` has `container_image(...)` as the last step in the `BUILD.bazel`.

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -81,6 +81,9 @@ jobs:
         wget -O- -nv https://github.com/google/ko/releases/download/v0.8.0/ko_0.8.0_Darwin_x86_64.tar.gz > /tmp/ko.tgz
         tar xzvf /tmp/ko.tgz -C /tmp/bin
 
+        wget -O- -nv https://github.com/bazelbuild/bazel/releases/download/4.2.0/bazel-4.2.0-darwin-x86_64 > /tmp/bin/bazel
+        chmod +x /tmp/bin/bazel
+
         cd "src/github.com/${{ github.repository }}"
 
         export KBLD_E2E_SKIP_WHEN_HTTP_REGISTRY=true
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -27,7 +27,7 @@ var (
 	// INFO: Build completed successfully, 1 total action
 	// Loaded image ID: sha256:328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e
 	// Tagging 328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e as bazel:simple-app
-	bazelImageID = regexp.MustCompile("Loaded image ID: (sha256:)?([0-9a-z]+)")
+	bazelImageID = regexp.MustCompile("Loaded image ID: (sha256:)([0-9a-z]+)")
 )
 
 type Bazel struct {
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -51,11 +51,11 @@ func (b *Bazel) Build(image, directory string, opts config.SourceBazelBuildOpts)
 
 		cmdArgs := []string{"run"}
 
-		if opts.Label == nil {
-			return ctlbdk.DockerTmpRef{}, fmt.Errorf("Expected label to be specified, but was not")
+		if opts.Target == nil {
+			return ctlbdk.DockerTmpRef{}, fmt.Errorf("Expected target to be specified, but was not")
 		}
 
-		cmdArgs = append(cmdArgs, *opts.Label)
+		cmdArgs = append(cmdArgs, *opts.Target)
 
 		if opts.RawOptions != nil {
 			cmdArgs = append(cmdArgs, *opts.RawOptions...)
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -30,7 +30,7 @@ func NewBazel(docker ctlbdk.Docker, logger ctllog.Logger) Bazel {
 	return Bazel{docker: docker, logger: logger}
 }
 
-func (b *Bazel) Run(image, directory string, opts config.SourceBazelBuildOpts) (ctlbdk.DockerTmpRef, error) {
+func (b *Bazel) Run(image, directory string, opts config.SourceBazelRunOpts) (ctlbdk.DockerTmpRef, error) {
 	prefixedLogger := b.logger.NewPrefixedWriter(image + " | ")
 
 	prefixedLogger.Write([]byte(fmt.Sprintf("starting build (using bazel): %s\n", directory)))
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -1,0 +1,85 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package image
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os/exec"
+	"regexp"
+
+	ctlbdk "github.com/k14s/kbld/pkg/kbld/builder/docker"
+	"github.com/k14s/kbld/pkg/kbld/config"
+	ctllog "github.com/k14s/kbld/pkg/kbld/logger"
+)
+
+var (
+	// Example output that includes the final digest:
+	// INFO: Analyzed target //:simple-app (0 packages loaded, 0 targets configured).
+	// INFO: Found 1 target...
+	// Target //:simple-app up-to-date:
+	//   bazel-bin/simple-app-layer.tar
+	// INFO: Elapsed time: 0.148s, Critical Path: 0.00s
+	// INFO: 1 process: 1 internal.
+	// INFO: Build completed successfully, 1 total action
+	// INFO: Build completed successfully, 1 total action
+	// Loaded image ID: sha256:328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e
+	// Tagging 328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e as bazel:simple-app
+	bazelImageID = regexp.MustCompile("Loaded image ID: (sha256:)?([0-9a-z]+)")
+)
+
+type Bazel struct {
+	docker ctlbdk.Docker
+	logger ctllog.Logger
+}
+
+func NewBazel(docker ctlbdk.Docker, logger ctllog.Logger) Bazel {
+	return Bazel{docker: docker, logger: logger}
+}
+
+func (b *Bazel) Build(image, directory string, opts config.SourceBazelBuildOpts) (ctlbdk.DockerTmpRef, error) {
+
+	prefixedLogger := b.logger.NewPrefixedWriter(image + " | ")
+
+	prefixedLogger.Write([]byte(fmt.Sprintf("starting build (using bazel): %s\n", directory)))
+	defer prefixedLogger.Write([]byte("finished build (using bazel)\n"))
+
+	var imageID string
+	{
+		var stdoutBuf, stderrBuf bytes.Buffer
+
+		cmdArgs := []string{"run"}
+
+		if opts.Label == nil {
+			return ctlbdk.DockerTmpRef{}, fmt.Errorf("Expected label to be specified, but was not")
+		}
+
+		cmdArgs = append(cmdArgs, *opts.Label)
+
+		if opts.RawOptions != nil {
+			cmdArgs = append(cmdArgs, *opts.RawOptions...)
+		}
+
+		cmd := exec.Command("bazel", cmdArgs...)
+		cmd.Dir = directory
+		cmd.Stdout = io.MultiWriter(&stdoutBuf, prefixedLogger)
+		cmd.Stderr = io.MultiWriter(&stderrBuf, prefixedLogger)
+
+		err := cmd.Run()
+		if err != nil {
+			prefixedLogger.Write([]byte(fmt.Sprintf("error: %s\n", err)))
+			return ctlbdk.DockerTmpRef{}, err
+		}
+
+		matches := bazelImageID.FindStringSubmatch(stdoutBuf.String())
+		if len(matches) != 3 {
+			return ctlbdk.DockerTmpRef{}, fmt.Errorf("Expected to find image ID in bazel output but did not")
+		}
+
+		imageID = "sha256:" + matches[2]
+	}
+
+	return b.docker.RetagStable(ctlbdk.NewDockerTmpRef(imageID), image, imageID, prefixedLogger)
+}
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -40,7 +40,6 @@ func NewBazel(docker ctlbdk.Docker, logger ctllog.Logger) Bazel {
 }
 
 func (b *Bazel) Build(image, directory string, opts config.SourceBazelBuildOpts) (ctlbdk.DockerTmpRef, error) {
-
 	prefixedLogger := b.logger.NewPrefixedWriter(image + " | ")
 
 	prefixedLogger.Write([]byte(fmt.Sprintf("starting build (using bazel): %s\n", directory)))
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -16,15 +16,6 @@ import (
 )
 
 var (
-	// Example output that includes the final digest:
-	// INFO: Analyzed target //:simple-app (0 packages loaded, 0 targets configured).
-	// INFO: Found 1 target...
-	// Target //:simple-app up-to-date:
-	//   bazel-bin/simple-app-layer.tar
-	// INFO: Elapsed time: 0.148s, Critical Path: 0.00s
-	// INFO: 1 process: 1 internal.
-	// INFO: Build completed successfully, 1 total action
-	// INFO: Build completed successfully, 1 total action
 	// Loaded image ID: sha256:328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e
 	// Tagging 328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e as bazel:simple-app
 	bazelImageID = regexp.MustCompile("Loaded image ID: (sha256:)([0-9a-z]+)")
--- a/pkg/kbld/builder/bazel/bazel.go
+++ b/pkg/kbld/builder/bazel/bazel.go
@@ -30,7 +30,7 @@ func NewBazel(docker ctlbdk.Docker, logger ctllog.Logger) Bazel {
 	return Bazel{docker: docker, logger: logger}
 }
 
-func (b *Bazel) Build(image, directory string, opts config.SourceBazelBuildOpts) (ctlbdk.DockerTmpRef, error) {
+func (b *Bazel) Run(image, directory string, opts config.SourceBazelBuildOpts) (ctlbdk.DockerTmpRef, error) {
 	prefixedLogger := b.logger.NewPrefixedWriter(image + " | ")
 
 	prefixedLogger.Write([]byte(fmt.Sprintf("starting build (using bazel): %s\n", directory)))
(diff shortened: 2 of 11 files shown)


pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,75 @@
...
+	"os/exec"
+	"regexp"
+
+	ctlbdk "github.com/k14s/kbld/pkg/kbld/builder/docker"
+	"github.com/k14s/kbld/pkg/kbld/config"
+	ctllog "github.com/k14s/kbld/pkg/kbld/logger"
+)
+
+var (
+	bazelImageID = regexp.MustCompile("Loaded image ID: (sha256:)?([0-9a-z]+)")
```
pivotaljohn: Would you mind inserting some sample output in a comment? This typically includes the line that contains the digest and some surrounding lines of context.

Including this makes it easier to maintain the code: having the concrete example makes it crystal clear just how selective the regular expression is, etc.

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,75 @@
...
+	{
+		var stdoutBuf, stderrBuf bytes.Buffer
+
+		if opts.Label == nil {
+			return ctlbdk.DockerTmpRef{}, fmt.Errorf("Expected label to be specified, but was not")
+		}
+
+		cmdArgs := []string{"run"}
+
+		if opts.Label != nil {
```
pivotaljohn: nit: the guard clause above makes this check technically unnecessary.

Consider pulling down that check to this point (i.e. after line 46) and removing the redundant `if`.

test/e2e/assets/simple-app/WORKSPACE
```diff
@@ -0,0 +1,33 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "io_bazel_rules_go",
+    sha256 = "08369b54a7cbe9348eea474e36c9bbb19d47101e8860cec75cbf1ccd4f749281",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.24.0/rules_go-v0.24.0.tar.gz",
+        "https://github.com/bazelbuild/rules_go/releases/download/v0.24.0/rules_go-v0.24.0.tar.gz",
```
pivotaljohn: Is the selection of 0.24.0 of the Go rules specific/deliberate? Or could we select the latest compatible version?

- [v0.24.14](https://github.com/bazelbuild/rules_go/releases/tag/v0.24.14) looks to be the obvious latest on the v0.24.x branch.
- the [Announcements](https://github.com/bazelbuild/rules_go#announcements) section in this repo describes [v0.27.0](https://github.com/bazelbuild/rules_go/releases/tag/v0.27.0) as
  > This will be the last release on the 0.24 branch. 0.27 and 0.25 are now the two supported branches.
- [v0.28.0](https://github.com/bazelbuild/rules_go/releases/tag/v0.28.0) has what appears to be a breaking change (i.e. requiring Bazel 4.0.x)

test/e2e/assets/simple-app/WORKSPACE
```diff
@@ -0,0 +1,33 @@
...
+
+load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
+go_rules_dependencies()
+go_register_toolchains()
+
+http_archive(
+    name = "io_bazel_rules_docker",
+    sha256 = "59d5b42ac315e7eadffa944e86e90c2990110a1c8075f1cd145f487e999d22b3",
+    strip_prefix = "rules_docker-0.17.0",
+    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.17.0/rules_docker-v0.17.0.tar.gz"],
```
pivotaljohn: Same question as above regarding the version of Docker rules.

Latest version appears to be [v0.19.0](https://github.com/bazelbuild/rules_docker/releases/tag/v0.19.0). Is the version compatible with your setup?

.github/workflows/test-gh.yml
```diff
@@ -65,6 +65,9 @@ jobs:
         wget -O- https://github.com/google/ko/releases/download/v0.8.0/ko_0.8.0_Darwin_x86_64.tar.gz > /tmp/ko.tgz
         tar xzvf /tmp/ko.tgz -C /tmp/bin
 
+        wget -O- https://github.com/bazelbuild/bazel/releases/download/4.2.0/bazel-4.2.0-darwin-x86_64 > /tmp/bin/bazel
```
pivotaljohn:
```suggestion
        wget -O- -nv https://github.com/bazelbuild/bazel/releases/download/4.2.0/bazel-4.2.0-darwin-x86_64 > /tmp/bin/bazel
```

In a parallel PR (#153), we're attempting to reduce the noise that comes from the progress output from `wget`. Mind including the `-nv` flag, here?

pivotaljohn (commented):
Thank you for your contribution, @cliuj! 🙏🏻  The more builders that `kbld` supports, the more valuable it becomes as a build orchestrator. 👍🏻 

Below are some comments, questions, and suggestions.

test/e2e/assets/simple-app/WORKSPACE
cliuj: Nothing deliberate, they were just something local at the time. I will update the versions accordingly.

.github/workflows/test-gh.yml
pivotaljohn: 🙏🏻

pivotaljohn (approved):
LGTM.

Thanks again @cliuj for contributing this integration. 🙏🏻

pkg/kbld/config/config_bazel.go
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+type SourceBazelOpts struct {
+	Build SourceBazelBuildOpts
+}
+
+type SourceBazelBuildOpts struct {
+	Label      *string   `json:"label"`
```
cppforlife: It seems like correct term is actually target? Any reason why this is called a label?

```
 INFO: Analyzed target //:simple-app (0 packages loaded, 0 targets configured).
```

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,85 @@
...
+		cmd.Stderr = io.MultiWriter(&stderrBuf, prefixedLogger)
+
+		err := cmd.Run()
+		if err != nil {
+			prefixedLogger.Write([]byte(fmt.Sprintf("error: %s\n", err)))
+			return ctlbdk.DockerTmpRef{}, err
+		}
+
+		matches := bazelImageID.FindStringSubmatch(stdoutBuf.String())
+		if len(matches) != 3 {
```
cppforlife, line 67: im guessing it's possible that bazel output can have multiple images loaded? i think we should fail instead of just picking up first one.

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,85 @@
...
+	docker ctlbdk.Docker
+	logger ctllog.Logger
+}
+
+func NewBazel(docker ctlbdk.Docker, logger ctllog.Logger) Bazel {
+	return Bazel{docker: docker, logger: logger}
+}
+
+func (b *Bazel) Build(image, directory string, opts config.SourceBazelBuildOpts) (ctlbdk.DockerTmpRef, error) {
+
```
cppforlife: nit: remove newline

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,85 @@
...
+	// INFO: Found 1 target...
+	// Target //:simple-app up-to-date:
+	//   bazel-bin/simple-app-layer.tar
+	// INFO: Elapsed time: 0.148s, Critical Path: 0.00s
+	// INFO: 1 process: 1 internal.
+	// INFO: Build completed successfully, 1 total action
+	// INFO: Build completed successfully, 1 total action
+	// Loaded image ID: sha256:328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e
+	// Tagging 328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e as bazel:simple-app
+	bazelImageID = regexp.MustCompile("Loaded image ID: (sha256:)?([0-9a-z]+)")
```
cppforlife: do we need `?` in `(sha256:)?`

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,85 @@
...
+
+		matches := bazelImageID.FindStringSubmatch(stdoutBuf.String())
+		if len(matches) != 3 {
+			return ctlbdk.DockerTmpRef{}, fmt.Errorf("Expected to find image ID in bazel output but did not")
+		}
+
+		imageID = "sha256:" + matches[2]
+	}
+
+	return b.docker.RetagStable(ctlbdk.NewDockerTmpRef(imageID), image, imageID, prefixedLogger)
```
cppforlife, line 74: hmm, is this actually succeeding? `ctlbdk.NewDockerTmpRef(imageID)` ends up to be `sha256:...` which i dont think is a valid tag docker sees? (or is docker allowing digest refs be the first arg to docker tag?)

pkg/kbld/config/config_bazel.go
cliuj: Misinterpretation on my part. Label syntax and target patterns syntax are almost similar and can be interchangeable. 

It seems that `target` is more appropriate here, so I'll change it to use `target`.

pkg/kbld/builder/bazel/bazel.go
cliuj: I'm not entirely sure of all the build outcomes for bazel, but in our use case there are multiple stages/layers in our bazel build where we do load an image which becomes the base image of the next stage/layer.

Something like:
```
path/to/project | Loaded image: bazel/image:image
```

However, the final stage output is always:
```
Loaded image ID: sha256:<digest>
```
which is what regex for here. The extra `ID` _seems_ unique enough for the final output.

pkg/kbld/builder/bazel/bazel.go
cliuj: Don't the tags get automatically cleaned with `"-"`? The output results in:

```
simple-app | Loaded image ID: sha256:9e5eeee36c61be12f84bfdff09a14805f33d7056c2dac8d6758d7bd04d25a424
simple-app | Tagging 9e5eeee36c61be12f84bfdff09a14805f33d7056c2dac8d6758d7bd04d25a424 as bazel:simple-app
simple-app | finished build (using bazel)
resolve | final: simple-app -> kbld:simple-app-sha256-9e5eeee36c61be12f84bfdff09a14805f33d7056c2dac8d6758d7bd04d25a424
```

This was based off of the `pack` builder's implementation.

pkg/kbld/builder/bazel/bazel.go
cliuj: Nope, will remove

pkg/kbld/builder/bazel/bazel.go
cppforlife: i was thinking if your bazel config contains something like this:

```
go_image(
    name = "simple-app-go-image",
    srcs = ["app.go"],
    importpath = "github.com/carvel-kbld/test/e2e/assets/simple-app",
)

go_image(
    name = "simple-app-go-image-2",
    srcs = ["app.go"],
    importpath = "github.com/carvel-kbld/test/e2e/assets/simple-app",
)
```

that are both leaf images.

pkg/kbld/builder/bazel/bazel.go
cppforlife: hmm, you are right that pack seems to be doing that as well.

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,84 @@
...
+	prefixedLogger := b.logger.NewPrefixedWriter(image + " | ")
+
+	prefixedLogger.Write([]byte(fmt.Sprintf("starting build (using bazel): %s\n", directory)))
+	defer prefixedLogger.Write([]byte("finished build (using bazel)\n"))
+
+	var imageID string
+	{
+		var stdoutBuf, stderrBuf bytes.Buffer
+
+		cmdArgs := []string{"run"}
```
pivotaljohn, line 43: Are we deliberately doing a "bazel [run](https://docs.bazel.build/versions/4.2.1/user-manual.html#run)" rather than a "bazel [build](https://docs.bazel.build/versions/4.2.1/guide.html#the-build-command)"?

pkg/kbld/builder/bazel/bazel.go
```diff
@@ -0,0 +1,84 @@
...
+	// INFO: Analyzed target //:simple-app (0 packages loaded, 0 targets configured).
+	// INFO: Found 1 target...
+	// Target //:simple-app up-to-date:
+	//   bazel-bin/simple-app-layer.tar
+	// INFO: Elapsed time: 0.148s, Critical Path: 0.00s
+	// INFO: 1 process: 1 internal.
+	// INFO: Build completed successfully, 1 total action
+	// INFO: Build completed successfully, 1 total action
+	// Loaded image ID: sha256:328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e
+	// Tagging 328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e as bazel:simple-app
```
pivotaljohn:
```suggestion
	// Loaded image ID: sha256:328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e
	// Tagging 328b5f47550c85cea5284911ad4d284ce20e8240d61d2610eb6cb4aa8b43c19e as bazel:simple-app
```

- bazel build output (i.e. what's being suggested to delete) is sent to stderr .
- `./bazel-bin/simple-app.executable` stdout; its output consists entirely of the two remaining lines ☝🏻 
- let's remove the stderr output as it is not subject to the output scraping we're performing within `Build()`.

pkg/kbld/builder/bazel/bazel.go
pivotaljohn: oh... I think I answered my own question.

"bazel build" produces (among other artifacts) a script that will config, (docker) load, and tag the image (e.g. `bazel-bin/simple-app.executable`), but that script is not executed.
"bazel run" executes the script. 👍🏻

pkg/kbld/builder/bazel/bazel.go
pivotaljohn: Note that:

`bazel build` builds _one or many_ targets:

```bash
$ bazel build :simple-app1 :simple-app2
INFO: Analyzed 2 targets (85 packages loaded, 7287 targets configured).
INFO: Found 2 targets...
INFO: Elapsed time: 4.869s, Critical Path: 3.13s
INFO: 68 processes: 24 internal, 44 darwin-sandbox.
INFO: Build completed successfully, 68 total actions
$ 
```

while `bazel run` builds and executes _exactly one_ target (and it's the first one mentioned):

```bash
$ bazel run :simple-app1 :simple-app2
INFO: Analyzed target //:simple-app1 (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:simple-app1 up-to-date:
  bazel-bin/simple-app1-layer.tar
INFO: Elapsed time: 0.171s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
8651333b21e7: Loading layer [==================================================>]  3.031MB/3.031MB
ba16d454860a: Loading layer [==================================================>]  15.44MB/15.44MB
ea465d14ddf7: Loading layer [==================================================>]  6.472MB/6.472MB
84ff92691f90: Loading layer [==================================================>]  10.24kB/10.24kB
Loaded image ID: sha256:9e5eeee36c61be12f84bfdff09a14805f33d7056c2dac8d6758d7bd04d25a424
Tagging 9e5eeee36c61be12f84bfdff09a14805f33d7056c2dac8d6758d7bd04d25a424 as bazel:simple-app1
$
```

Note that `bazel build` produces all of the artifacts (i.e. tarball, OCI image config, digests); it's that they are not loaded into the local image repo.

Will it be explicit/obvious to a user that the Bazel integration with `kbld` is performing a `bazel run` (rather than a `bazel build`)?

cliuj: >Will it be explicit/obvious to a user that the Bazel integration with kbld is performing a bazel run (rather than a bazel build)?

This was something I was worried about as well. `bazel run`was used here becase it was the only way to load the image into the local image repo, but `bazel run <target>` isn't limited to just image building (it can build and execute the built image if the last stage was something like `go_image(...)`). The final step that seemed safe was `container_image(...)` because it just loads the images, but does not execute them.

I'm not sure if this would be obvious to a user. Perhaps we can add a note here: https://carvel.dev/kbld/docs/latest/config/#sources about the underlying command being `bazel run` instead of `bazel build` and using `container_image(...)`.

pkg/kbld/builder/bazel/bazel.go
pivotaljohn: 🙏🏻

pkg/kbld/builder/bazel/bazel.go
cppforlife: for docker source (https://carvel.dev/kbld/docs/latest/config/#docker) we have explicit build subsection. might make sense to intro this for bazel (e.g. bazel.run section)

edit: oh looks like we already have build section under bazel, i think we should rename it to run.

cppforlife: > while bazel run builds and executes exactly one target (and it's the first one mentioned):

the example i was providing with multiple container_images would still be considered a single target right?

cliuj: > > while bazel run builds and executes exactly one target (and it's the first one mentioned):
> 
> the example i was providing with multiple container_images would still be considered a single target right?

Yup, in this case, target can only be either `simple-app-go-image` or `simple-app-go-image-2`, and the output will be the one chosen.


> for docker source (https://carvel.dev/kbld/docs/latest/config/#docker) we have explicit build subsection. might make sense to intro this for bazel (e.g. bazel.run section)
> 
> edit: oh looks like we already have build section under bazel, i think we should rename it to run.

I can change it to `run`. I'm assuming something like:
```
  bazel:
    # instead of "build:"
    run:
      target: :simple-app
```

pkg/kbld/config/config_bazel.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+type SourceBazelOpts struct {
+	Run SourceBazelBuildOpts
```
cppforlife: small nit: SourceBazelBuildOpts -> SourceBazelRunOpts

pkg/kbld/builder/bazel/bazel.go
cppforlife: got it. yup on run.

cppforlife:
(failure unrelated to this PR, @jtigger is chasing it down)

getting really close!

pivotaljohn:
> (failure unrelated to this PR, @jtigger is chasing it down)
> 
> getting really close!

Yup yup; we recently added a step in the workflow to authenticate to DockerHub to avoid rate limiting.  GitHub (appropriately) prevents Actions running off a fork of the repo from accessing the `secrets` context.  We didn't account for that fact in the script: the secret is empty, the script attempts to login with an empty input to stdin, and `docker` gripes and errors because the only alternative would be for the tty to be interactive (which it is not).

For now, I'll short-circuit that authentication if secrets are not provided. Pull requests from forks will run unauthenticated.

Tracking a longer-term solution here: vmware-tanzu/carvel-kbld#159

pivotaljohn:
Pedantically, fixing the build for this PR would require a rebase.
Practically, I pulled and ran the tests locally (including https://github.com/vmware-tanzu/carvel-kbld/pull/149/commits/e4c5ee8b760b9a37fc45ca68121556f3ecfd2c0e); all ran green. 👍🏻 

@cliuj — would you do me a favor and review the "Bazel" section of the proposed docs?
- Preview -- https://deploy-preview-234--carvel.netlify.app/kbld/docs/latest/config/#bazel
- PR -- https://github.com/vmware-tanzu/carvel/pull/234/files#diff-a942671466ca8d61b968f6cc660959f1f6fcdd5468b940c224e5f12a438c537bR415

PR #153, opened 2021-09-01
Add DockerHub credentials to E2E in CI

### LINKED ISSUE #146: GitHub worklow tests should test latest (or a more recent) version of minikube/k8s
**Describe the problem/challenge you have**
I noticed a [kapp PR](https://github.com/vmware-tanzu/carvel-kapp/pull/274/files) was submitted to test against the latest version of minikube, and therefore a recent version of kubernetes. I noticed that [kbld is testing against an outdated version of minikube/k8s](https://github.com/vmware-tanzu/carvel-kbld/blob/404cca83c1c7a1775d5fe09940d1641058075ce7/.github/workflows/test-gh.yml#L45). This creates a reduced sense of confidence in kbld's ability to work with the latest versions of kubernetes.

**Describe the solution you'd like**
An automated way for kbld to test against the latest, or at least a recent version, of kubernetes.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

pivotaljohn:
In order to avoid rate limiting errors, uses injected credentials:
- authenticate the docker client;
- authenticate image pulls for Pods deployed as part of the test (today, just the `registry2`);
- provide credentials for kubectl-buildkit.

Includes minor improvements:
- reduce output from `wget` calls to reduce that noise
- remove some dead lines
- resolves #146

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -55,11 +55,11 @@ jobs:
         # ... for builders using Docker
         echo "${DOCKERHUB_ACCESS_TOKEN}" | docker login --username "${DOCKERHUB_USERNAME}" --password-stdin
         # ... for kubectl-buildkit
-        kubectl create secret docker-registry buildkit \
+        minikube kubectl --ssh=true -- create secret docker-registry buildkit \
           --docker-username="${DOCKERHUB_USERNAME}" \
           --docker-password="${DOCKERHUB_ACCESS_TOKEN}"
         # ... for any other service account that pulls images from DockerHub
-        kubectl create secret docker-registry dockerhub-credentials \
+        minikube kubectl --ssh=true -- create secret docker-registry dockerhub-credentials \
           --docker-username="${DOCKERHUB_USERNAME}" \
           --docker-password="${DOCKERHUB_ACCESS_TOKEN}"
 
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -37,7 +37,8 @@ jobs:
         set -e -x
 
         mkdir -p ~/.docker/machine/cache
-        curl -Lo ~/.docker/machine/cache/boot2docker.iso https://github.com/boot2docker/boot2docker/releases/download/v19.03.12/boot2docker.iso
+        wget -O -nv https://github.com/boot2docker/boot2docker/releases/download/v19.03.12/boot2docker.iso > ~/.docker/machine/cache/boot2docker.iso
+
         brew install docker docker-machine
 
         # Install ytt, kapp for build and tests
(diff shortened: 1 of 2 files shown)


.github/workflows/test-gh.yml
```diff
@@ -29,40 +30,53 @@ jobs:
         path: src/github.com/${{ github.repository }}
         fetch-depth: 0
     - name: Run Tests
+      env:
+        DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
```
cppforlife: @pivotaljohn what account are we including creds for?

pivotaljohn: Today, it's "k8slt".

.github/workflows/test-gh.yml
```diff
@@ -29,40 +30,53 @@ jobs:
...
-        wget -O- https://github.com/kubernetes/minikube/releases/download/v1.10.0/minikube-darwin-amd64 > /tmp/bin/minikube
+        wget -O- -nv https://github.com/kubernetes/minikube/releases/latest/download/minikube-darwin-amd64 > /tmp/bin/minikube
         chmod +x /tmp/bin/minikube
-        minikube start --driver=virtualbox --insecure-registry=192.168.99.100/16
+        minikube start --driver=virtualbox --insecure-registry=192.168.0.0/16
         eval $(minikube docker-env --shell=bash)
 
+        # Authenticate to DockerHub to avoid test failures due to rate limiting
+        # ... for builders using Docker
+        echo "${DOCKERHUB_ACCESS_TOKEN}" | docker login --username "${DOCKERHUB_USERNAME}" --password-stdin
```
DennisDenuto: With the above bash option`set -x` I think bash will print out the eval'd `DOCKERHUB_ACCESS_TOKEN` variable. 

for e.g.
```
bash-5.0$ FOO=test
bash-5.0$ set -x
bash-5.0$ echo $FOO | grep bar
+ echo test
+ grep bar
bash-5.0$
```

But I'm pretty sure [github actions will redact this](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-secrets). But!... If you wanted to be safer than sorry you could wrap this around `set +x / set -x`, or potentially drop the `set -x` on top? (it should make the logs less verbose, hopefully not reducing the ease of debugging by much either).

.github/workflows/test-gh.yml
pivotaljohn: Updated so that the username is in plaintext. 👍🏻

.github/workflows/test-gh.yml
pivotaljohn: Yeah, it _is_ redacted, but not interested in relying on that. I'll wrap the call as it's particularly helpful in troubleshooting to have `set -x`... and we've done _other_ work to significantly reduce the total log size. 👍🏻

DennisDenuto: that makes sense! i'll resolve this convo :-)

pivotaljohn:
@cppforlife + @DennisDenuto : GTG?

PR #115, opened 2021-06-22
Update golang crypto library

joaopapereira:

PR #154, opened 2021-09-02
Upgrade golang crypto library

gcheadle-vmware:


cppforlife:
@gcheadle-vmware no files have changed in the vendor directory... also why do we need indirect dependency on this.

gcheadle-vmware:
This PR was created as a replacement for #115, the changes in #115 were already in develop (likely through a dependabot PR). I was not aware of the motivation behind manually bumping the crypto library, but I wanted to try it myself for practice. As a follow up, @DennisDenuto and I locally updated all dependancies and did not see a bump for the crypto lib, so we thought it would be best to close this PR and allow dependabot to take care of bumps in the future.

Issue #95, opened 2021-01-25
kbld inspect should have an output option for lock file


voor:
**Describe the problem/challenge you have**
I am bringing in new capabilities, for example from a `helm template` call, and I want to quickly and efficiently build a kbld configuration file that I know encompasses the images outlined in that template.

```
$ helm template harbor bitnami/harbor | kbld inspect -f -
Images

Image     docker.io/bitnami/chartmuseum:0.12.0-debian-10-r261  
Metadata  -  
Resource  deployment/harbor-chartmuseum (apps/v1) cluster  

... [removed for brevity]

Image     docker.io/bitnami/redis:6.0.9-debian-10-r66  
Metadata  -  
Resource  statefulset/harbor-redis-master (apps/v1) namespace: default  

14 images

Succeeded
```

This output, while very helpful, requires a fair bit of massaging to get into a kbld config format.  There is a JSON output, but even that output is a far cry from kbld's config format.

**Describe the solution you'd like**
Something along these lines:

```
$ helm template harbor bitnami/harbor | kbld inspect -f - -o kbld
apiVersion: kbld.k14s.io/v1alpha1
kind: Config
minimumRequiredVersion: 0.24.0
overrides:
- image: docker.io/bitnami/chartmuseum:0.12.0-debian-10-r261
- image: docker.io/bitnami/harbor-adapter-clair:2.1.2-debian-10-r34
- image: docker.io/bitnami/harbor-adapter-trivy:2.1.2-debian-10-r37
- image: docker.io/bitnami/harbor-clair:2.1.2-debian-10-r33
- image: docker.io/bitnami/harbor-core:2.1.2-debian-10-r30
- image: docker.io/bitnami/harbor-jobservice:2.1.2-debian-10-r31
- image: docker.io/bitnami/harbor-notary-server:2.1.2-debian-10-r30
- image: docker.io/bitnami/harbor-notary-signer:2.1.2-debian-10-r31
- image: docker.io/bitnami/harbor-portal:2.1.3-debian-10-r0
- image: docker.io/bitnami/harbor-registry:2.1.2-debian-10-r31
- image: docker.io/bitnami/harbor-registryctl:2.1.2-debian-10-r31
- image: docker.io/bitnami/nginx:1.19.6-debian-10-r25
- image: docker.io/bitnami/postgresql:11.10.0-debian-10-r52
- image: docker.io/bitnami/redis:6.0.9-debian-10-r66
```

[...]

PR #156, opened 2021-09-03
add `--unprocessed-images-inspect` flag

pivotaljohn:
Sketch for conversation: https://kubernetes.slack.com/archives/CH8KCCKA5/p1630636823106500

### DIFF
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -26,14 +26,15 @@ import (
 type ResolveOptions struct {
 	ui ui.UI
 
-	FileFlags        FileFlags
-	RegistryFlags    RegistryFlags
-	AllowedToBuild   bool
-	BuildConcurrency int
-	ImagesAnnotation bool
-	ImageMapFile     string
-	LockOutput       string
-	ImgpkgLockOutput string
+	FileFlags         FileFlags
+	RegistryFlags     RegistryFlags
+	AllowedToBuild    bool
+	BuildConcurrency  int
+	ImagesAnnotation  bool
+	ImageMapFile      string
+	LockOutput        string
+	ImgpkgLockOutput  string
+	UnresolvedInspect bool
 }
 
 func NewResolveOptions(ui ui.UI) *ResolveOptions {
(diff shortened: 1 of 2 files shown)


pkg/kbld/cmd/resolve.go
```diff
@@ -78,7 +80,20 @@ func (o *ResolveOptions) Run() error {
 
 	imgFactory := ctlimg.NewFactory(conf, registry, logger)
 
-	resolvedImages, err := o.resolveImages(nonConfigRs, conf, imgFactory)
+	imageURLs, err := o.collectImageReferences(nonConfigRs, conf)
+	if err != nil {
+		return err
+	}
+	if o.UnprocessedImageInspect {
```
cppforlife: nit: add newline before

pkg/kbld/cmd/unprocessed_image_urls.go
```diff
@@ -33,3 +35,7 @@ func (i *UnprocessedImageURLs) All() []UnprocessedImageURL {
 	})
 	return result
 }
+
+func (i *UnprocessedImageURLs) Bytes() ([]byte, error) {
```
cppforlife, line 39: dont think we want to use `url: ...` in its output. prolly better to go with `image: ...` (locking down serialized format independently from UnprocessedImageURL struct might be worth it?)

cppforlife:
any thoughts on naming and output for processed images?

pkg/kbld/cmd/unprocessed_image_urls.go
pivotaljohn: I think, for now, placing the json struct tags on the internal data structure is _sufficient_ for now. Might be something to revisit as other requests (like #95) are addressed.

pivotaljohn:
> any thoughts on naming and output for processed images?

Updated per suggestions. E2E tests are running (takes ~30min); these changes should have no affect on them.

pkg/kbld/cmd/unprocessed_image_urls.go
```diff
@@ -5,14 +5,16 @@ package cmd
...
 )
 
 type UnprocessedImageURL struct {
-	URL string
+	URL string `json:"image"`
 }
 
 type UnprocessedImageURLs struct {
-	urls map[UnprocessedImageURL]struct{}
+	urls map[UnprocessedImageURL]struct{} `json:"unresolved"`
```
cppforlife, line 17: hmm, ytt doesnt have map key as top level when doing --files-inspect/--data-values-inspect. but its also probably a good idea to have it... should we rename this to `images`?

pivotaljohn: Chose "unresolved" to make clear the status. But, I can see more useful if the data structure were consistent (e.g. if/when another inspect is inserted later in the processing).

Whether a top-level _should_ be present: is it possible we'd like to output a mixture of objects as "current state" in this or subsequent inspects? If yes, then the top-level key is needed; if not, then its presence is somewhat dubious.

PR #163, opened 2021-09-17
Reorganize "meta" fields

pivotaljohn:
- remove "yaml:" struct field tags; our YAML is being processed through
  sigs.k8s.io/yaml (which performs (un)marshaling through the JSON
  parser, not directly through go-yaml/yaml.
- lowercase all field names (consistent with YAML around it, making it easier to read)
- promote "Type" fields as top-level keys for each type of meta.

### DIFF
--- a/pkg/kbld/config/image_meta.go
+++ b/pkg/kbld/config/image_meta.go
@@ -14,7 +14,7 @@ type Meta interface {
 }
 
 type imageMeta struct {
-	Metas []Meta
+	Metas []Meta `json:"metas,omitempty"`
 }
 
 const (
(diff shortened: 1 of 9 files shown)


pkg/kbld/config/image_meta.go
```diff
@@ -26,33 +26,33 @@ const (
 )
 
 type BuiltImageSourceGit struct {
-	Type      string // always set to GitMeta
-	RemoteURL string `json:",omitempty" yaml:",omitempty"`
-	SHA       string
-	Dirty     bool
-	Tags      []string `json:",omitempty" yaml:",omitempty"`
+	Type      string   `json:"type"` // always set to GitMeta
+	RemoteURL string   `json:"remoteUrl"`
```
cppforlife: here URL should still be upper case.

cppforlife:
since we are changing field names... should we just go ahead and change to something like this:

```
metas:
- git:
    remoteURL ...
- git:
    remoteURL: ...
- preresolved:
    ...
```

pivotaljohn:
> since we are changing field names... should we just go ahead and change to something like this:
> 
> ```
> metas:
> - git:
>     remoteURL ...
> - git:
>     remoteURL: ...
> - preresolved:
>     ...
> ```

❤️  — promoting "type:" as a top-level key fixes a complaint I have about it being difficult to visually parse this YAML given that the processing (through K8s' YAML parser marshaling/unmarshaling to/from the JSON parser) sorts the keys alphabetically.

We're breaking this "API", let's do it up right. 👍🏻

pkg/kbld/config/image_meta.go
pivotaljohn: done.

cppforlife:
superseded by #169

PR #164, opened 2021-09-18
allow disabling of building of images

cppforlife:
might be useful for context where only resolution should be allowed

### DIFF
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -27,6 +27,7 @@ type ResolveOptions struct {
 
 	FileFlags        FileFlags
 	RegistryFlags    RegistryFlags
+	AllowedToBuild   bool
 	BuildConcurrency int
 	ImagesAnnotation bool
 	ImageMapFile     string
@@ -46,6 +47,7 @@ func NewResolveCmd(o *ResolveOptions) *cobra.Command {
 	}
 	o.FileFlags.Set(cmd)
 	o.RegistryFlags.Set(cmd)
+	cmd.Flags().BoolVar(&o.AllowedToBuild, "build", true, "Allow building of images")
 	cmd.Flags().IntVar(&o.BuildConcurrency, "build-concurrency", 4, "Set maximum number of concurrent builds")
 	cmd.Flags().BoolVar(&o.ImagesAnnotation, "images-annotation", true, "Annotate resources with images annotation")
 	cmd.Flags().StringVar(&o.ImageMapFile, "image-map-file", "", "Set image map file (/cnab/app/relocation-mapping.json in CNAB)")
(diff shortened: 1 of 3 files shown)


pkg/kbld/cmd/resolve.go
```diff
@@ -27,6 +27,7 @@ type ResolveOptions struct {
 
 	FileFlags        FileFlags
 	RegistryFlags    RegistryFlags
+	AllowedToBuild   bool
```
pivotaljohn, line 30: Out of curiosity, was it coin toss or did you intentionally go with the affirmative flag name (e.g. as opposed to for example, `DisallowBuilds`).

pkg/kbld/image/err.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package image
+
+type ErrImage struct {
+	err error
+}
+
+var _ Image = ErrImage{}
```
pivotaljohn, line 10: Dig the "Null Object" pattern variant: "Error Object". ⭐ 

I would have likely included an `error` in the `BuiltImage`, myself.

But this is a great example of how to keep the overall system simpler by cordoning off a non-domain case (i.e. where a given resolution is destined to fail, independent of type: an "Erroring Image") into its own type (i.e. `ErrImage`).

One can completely understand `BuiltImage` without having to consider this case at all... it sits _just_ outside the sphere of the domain.

pkg/kbld/image/factory.go
```diff
@@ -43,6 +50,10 @@ func (f Factory) New(url string) Image {
 	}
 
 	if srcConf, found := f.shouldBuild(url); found {
+		if !f.opts.AllowedToBuild {
+			return NewErrImage(fmt.Errorf("Building of images is disallowed (tried to build '%s')", url))
```
pivotaljohn: 🤔  What are your thoughts on explicitly naming the mechanism that causes this to occur?  Like...

"Building of images is disallowed (tried to build '%s' because a source was configured for it)"

Being explicit about causes makes the error message accessible/actionable for even casual users.

I see one concern: that this function should not _assume_ why the build was attempted. So then, `shouldBuild()` should return a `string` that's the reason for why the build should occur...

```go
	if srcConf, found, reason := f.shouldBuild(url); found {
		if !f.opts.AllowedToBuild {
			return NewErrImage(fmt.Errorf("Building of images is disallowed (tried to build '%s' because %s)", url, reason))
```

pkg/kbld/cmd/resolve.go
cppforlife: the flag itself is called "build". internal name is AllowedToBuild, which probably has to make sense for the default value of true.

pkg/kbld/image/err.go
cppforlife: it's not a common pattern i use, but seems like it did ok job here.

pkg/kbld/image/factory.go
cppforlife: `reason = "a source was configured for it"` for now would effectively be hard coding. i would side on less work and just hard code into the error directly for now. unless you are thinking there are other reason that exist...

pivotaljohn: Going further takes us into a "purist" mindset. We're more pragmatic than that, here.

Let's hardcode this reason... it's — in fact — not likely to change.

PR #167, opened 2021-09-21
Rename "meta" to "origin"

pivotaljohn:
All of the information being stored in the "metas" section are
information about the sourcing of each image. The term "source" however,
is already used in this domain to mean a configuration from which an
image is produced from sources.

Origin is a softer claim than provenance; the latter of which is
generally taken to mean provable.


cppforlife:
superseded by https://github.com/vmware-tanzu/carvel-kbld/pull/169

PR #187, opened 2021-11-02
Add details for running the end to end test suite.

seemiller:
The README for the end to end test suite does not adequately prepare a contributor for running said suite. This adds the utilities that must be installed as a prerequisite and describes some environment variables that must/may be set.

Signed-off-by: Nicholas Seemiller <nseemiller@vmware.com>


joaopapereira (approved):
Looks good to me

PR #202, opened 2021-12-16
Change the module to match github repository

joaopapereira:


DennisDenuto:
LGTM

- Searched for 'k14s/kbld' in the codebase and couldn't find any references
- It builds locally fine `./hack/build.sh` && `./hack/build-binaries.sh`

It's worth considering calling this out as a breaking change for folk depending on kbld as a library.

PR #234, opened 2022-04-17
support platform selection flag and configuration

cppforlife:
allows users to select images (from img indexes) matching only specific platform via global flag or explicit configuration in overrides

### DIFF
--- a/pkg/kbld/cmd/resolve.go
+++ b/pkg/kbld/cmd/resolve.go
@@ -36,6 +36,7 @@ type ResolveOptions struct {
 	LockOutput        string
 	ImgpkgLockOutput  string
 	UnresolvedInspect bool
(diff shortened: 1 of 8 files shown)


voor:
Is there a global option to just have a kbld config that defaults every image to platformSelection?

cppforlife:
@voor `--platform` command line flag would do it. did you want config option as well?

voor:
We're trying to enable consistency without baking cli options into our scripts, but since we're on a single architecture this might be preemptive optimization

cppforlife:
i see, im a little bit hesitant to add a config option right now without seeing some concrete usage first (of flags or individual per image platform selection). a bit fuzzy if folks may have images with windows and linux mixed together, etc.

PR #241, opened 2022-05-09
Don't gpg sign test commits

benmoss:
Makes it pretty annoying if you have git configured to always GPG sign your commits, I didn't expect unit tests to ask for my GPG password


joaopapereira (approved):
LGTM

PR #244, opened 2022-05-11
support docker buildx builder

cppforlife:


hack/test-all-minikube-local-registry.sh
```diff
@@ -7,6 +7,30 @@ export KBLD_E2E_DOCKERHUB_HOSTNAME=$(minikube ip):30777
 # uncomment to disable stress tests
 # export KBLD_E2E_SKIP_STRESS_TESTS=true
 
-kapp deploy -a reg -f test/e2e/assets/minikube-local-registry.yml -y
+# Install secretgen to generate registry certs
+kapp deploy -a sg -f https://github.com/vmware-tanzu/carvel-secretgen-controller/releases/download/v0.8.0/release.yml -y
+# Install local docker2 registry
+kapp deploy -a reg -f <(ytt -f test/e2e/assets/minikube-local-registry.yml -v registry_alt_name=$(minikube ip)) -y
+
+if [ ! -z "$GITHUB_RUN_ID" ]; then
```
DennisDenuto: Since the bash unbounded option is enabled above on line 2 `set -u` and most developers will *not* have `GITHUB_RUN_ID` env var set, running this script locally on a dev machine I think will fail.

hack/test-all-minikube-local-registry.sh
```diff
@@ -7,6 +7,30 @@ export KBLD_E2E_DOCKERHUB_HOSTNAME=$(minikube ip):30777
...
+  # Install registry ca cert on the host machine
+  kubectl get secret registry-ca-cert -ojsonpath='{.data.crt\.pem}' | base64 --decode > registry-ca-cert.crt
```
DennisDenuto: I'm not sure why we are adding ca-certs to the registry? Or how this is tied to "supporting docker buildx builder" Esp since it looks like docker buildx is configured to insecure (http).

test/e2e/build_dockerbx_test.go
```diff
@@ -0,0 +1,105 @@
...
+- image: docker.io/*username*/kbld-e2e-tests-build
+  path: assets/simple-app
+  docker:
+    buildx: {}
+- image: docker.io/*username*/kbld-e2e-tests-build2
+  path: assets/simple-app
+  docker:
+    buildx:
+      # try out multi platform build
+      rawOptions: ["--platform=linux/amd64,linux/arm64,linux/arm/v7"]
```
DennisDenuto, line 79: It doesn't appear that this test makes any assertions that these platforms were in fact built.

Is this intentional / you don't want to test the underlying tool buildx itself?

I'm assuming you have manually tested passing rawOptions to buildx, and that it threads through correctly. (It looks like it should based off looking at the impl)

hack/test-all-minikube-local-registry.sh
cppforlife: this is coming from a commit that switches to linux, so this setup is for other builders. i believe it was bazel that needs that ca-cert configuration since its executing directly on the machine.

hack/test-all-minikube-local-registry.sh
cppforlife: yup, i actually run into this myself... forgot to fix this.

test/e2e/build_dockerbx_test.go
cppforlife: > Is this intentional / you don't want to test the underlying tool buildx itself?

yeah, ive tried to avoid testing the tool functionality itself. just testing that it built something. any thoughts how far we should go?

hack/test-all-minikube-local-registry.sh
DennisDenuto: Ah ok I see. So without adding this the bazel tests (or maybe some other builder test) will fail due to unable to verify the registry cert.

It also looks like the registry previously wasn't configured with TLS. This change will result in our builder tests running against a more 'realistic' registry configuration out in the wild. That is a nice plus :-)

DennisDenuto: However, on the other hand, I'm running into issues when running tests that push / interact with the registry since the registry ca-cert is not provided to the client.

Specifically running `TestDockerBuildxBuildAndPushSuccessful` locally results in `error: Get "https://192.168.59.100:30777/v2/": x509: certificate signed by unknown authority`

test/e2e/build_dockerbx_test.go
DennisDenuto: I think as far as e2e tests go, the boundary you picked makes sense to me. It is testing building an image and pushing the built image which is the main use case. (Although the push test fails locally for me due to missing certs on my host machine; [see above comment](https://github.com/vmware-tanzu/carvel-kbld/pull/244/files#r883026849))

However If *you* are concerned about catching bugs/regression (such as ensuring that the `rawOptions` is correctly threaded into the underlying tool) I think a 'unit test' / 'integration test' could be added. 

fwiw the tests introduced in this PR seem enough to me.

hack/test-all-minikube-local-registry.sh
benmoss:
```suggestion
if [ -n "${GITHUB_RUN_ID-}" ]; then
```

hack/test-all-minikube-local-registry.sh
benmoss: Yeah, I'm also seeing this failure locally

test/e2e/build_dockerbx_test.go
benmoss: Yeah I ran into a bug with the [kubectl-buildkit](https://github.com/vmware-tanzu/buildkit-cli-for-kubectl) implementation in kbld that it wouldn't actually give you the digest for the multi-arch image, even if you passed those args correctly. I never looked into the bug. It seems to work correctly with this implementation, but it is a possibility apparently.

hack/test-all-minikube-local-registry.sh
DennisDenuto: Now that tests are [run within a container](https://github.com/vmware-tanzu/carvel-kbld/blob/develop/hack/test-all-locally.sh) (populated with all the dependencies required), It *should* be possible to remove this `if [ ! -z "$GITHUB_RUN_ID" ]; then` statement, since the 'test container' is based off a debian image. 

this should result in always populating ca-certs regardless of whether tests were run in gh-action or locally on a dev workstation, since *both* use the same test container to run all the tests.

test/e2e/build_dockerbx_test.go
cppforlife: > fwiw the tests introduced in this PR seem enough to me.

all right, can possibly intro additional tests in additional PRs.

PR #252, opened 2022-05-25
Prevent appending duplicate preresolved origins

DennisDenuto:
Notes:

- Mainly using this branch to get faster feedback running *all* the tests in gh-runner
- Only add a Preresolved image origin if it doesn't already exist. This avoids adding duplicate Preresolved images.
- Stumbled & Fixes on I *think* a bug around adding additional config to an Image Map config. 
- refactor out the test dependency executables into a separate OCI. Stops requiring a dev to install older versions of a tool on their host to run e2e tests. Also allows our github test action to re-use this.

Note to reviewer:
- there are some conflicting changes between this PR and https://github.com/vmware-tanzu/carvel-kbld/pull/244
specifically: both PR's make non-minor changes to `test-gh.yml`. These non-minor changes seem complimentary but a merge conflict will probably occur.

Authored-by: Dennis Leon <leonde@vmware.com>


pkg/kbld/image/preresolved.go
```diff
@@ -17,6 +17,13 @@ func NewPreresolvedImage(url string, origins []ctlconf.Origin) PreresolvedImage
 }
 
 func (i PreresolvedImage) URL() (string, []ctlconf.Origin, error) {
+	for _, origin := range i.origins {
+		if origin.Preresolved != nil && origin.Preresolved.URL == i.url {
+			imageOrigins := copyAndAppendOrigins(i.origins)
+			return i.url, imageOrigins, nil
+		}
+	}
```
cppforlife, line 25: what if we do a more generic check for any dups (exact match) within metas?

DennisDenuto: Spoke over zoom and it will be simpler to not make a generic solution due to this bug only manifesting for preresolved images

PR #339, opened 2023-03-02
Set environment on tests to use the correct secrets

joaopapereira:


joaopapereira:
Another PR fixed the issue this PR is trying to fix

PR #354, opened 2023-04-02
Replacing metric-proxy stale image in v0.35.x line

sethiyash:
- Replacing metric-proxy stale image in v0.35.x line


joaopapereira (approved):
LGTM

PR #386, opened 2023-06-15
Bump go 1.20.5 in line  0.35.x

kumaritanushree:


hack/Dockerfile.dev
```diff
@@ -1,7 +1,7 @@
-FROM golang:1.19
+FROM golang:1.20
 
 RUN apt-get update -y
-RUN apt-get install docker.io apt-transport-https ca-certificates gnupg python -y
+RUN apt-get install docker.io apt-transport-https ca-certificates gnupg python-is-python3 -y
```
praveenrewar, line 4: Don't we want to make this change in develop first?

praveenrewar (commented):
Quick question ^

hack/Dockerfile.dev
joaopapereira: The change was already merged in develop

joaopapereira (approved):
LGTM

PR #390, opened 2023-06-19
Updated imgpkg and vendir in 0.35.x

kumaritanushree:


kumaritanushree:
No need to update these as we are not seeing any CVE for these.

PR #411, opened 2023-09-12
chore: remove refs to deprecated io/ioutil

testwill:


cmd/kbld/kbld.go
```diff
@@ -18,7 +18,7 @@ import (
 func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
 
-	log.SetOutput(ioutil.Discard)
+	log.SetOutput(io.Discard)
```
sethiyash, line 21: Here also should we use the `os` package only as it is already imported so we can ignore importing `io`?

pkg/kbld/resources/file_sources.go
```diff
@@ -24,7 +24,7 @@ func NewStdinSource() StdinSource { return StdinSource{} }
 func (s StdinSource) Description() string { return "stdin" }
 
 func (s StdinSource) Bytes() ([]byte, error) {
-	return ioutil.ReadAll(os.Stdin)
+	return io.ReadAll(os.Stdin)
```
sethiyash, line 27: Here also should we use the `os` package only as it is already imported so we can ignore importing `io`?

pkg/kbld/resources/file_sources.go
```diff
@@ -63,7 +63,7 @@ func (s HTTPFileSource) Bytes() ([]byte, error) {
 
 	defer resp.Body.Close()
 
-	result, err := ioutil.ReadAll(resp.Body)
+	result, err := io.ReadAll(resp.Body)
```
sethiyash, line 66: Here also should we use the `os` package only as it is already imported so we can ignore importing `io`?

sethiyash:
Hi @testwill,
Thanks for raising this PR. This makes total sense to remove io/ioutil as it is deprecated as of Go 1.16.

pkg/kbld/resources/file_sources.go
testwill: os.ReadAll? There is no such function

sethiyash: oh! my bad. We are good then.

sethiyash (approved):

joaopapereira (approved):
LGTM

PR #412, opened 2023-09-15
Bumping go version to 1.21.1

sethiyash:
- Bumping go version to 1.21.1


praveenrewar:
@sethiyash I am guessing we also need to update hack/Dockerfile.dev?

joaopapereira (approved):
LGTM

PR #423, opened 2023-10-17
Remove tools cache in github runner

praveenrewar:


.github/workflows/test-gh.yml
```diff
@@ -20,6 +20,12 @@ jobs:
     runs-on: ubuntu-latest
     environment: DockerHub E2E
     steps:
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
```
joaopapereira: Can we change this to be a specific sha? just to make sure we are always using the same version.

joaopapereira (changes requested):
Let us change the current action from the branch to a specific SHA to ensure we always use the same version.

.github/workflows/test-gh.yml
praveenrewar: Switched to using the latest released version instead of main.

joaopapereira (approved):
LGTM

PR #457, opened 2024-02-15
Change registry query from GET to HEAD to avoid Dockerhub API rate li…

antwacky:
…mits


joaopapereira (changes requested):
Unfortunately, it needs to be a little bit more complicated than that. Some registries  do not like when you send Head requests in some situations, so in order to ensure that these still work we need to adopt the same logic we currently have [in imgpkg](https://github.com/carvel-dev/imgpkg/blob/develop/pkg/imgpkg/registry/registry.go#L358)

antwacky:
Ah thanks for the pointer, I've updated to use HEAD in the first instance and fall back to GET in case of error as pointed out in the imgpkg library.

joaopapereira (approved):
LGTM

PR #458, opened 2024-02-21
Add error message when checking if file exists

joaopapereira:


pkg/kbld/resources/file_resources.go
```diff
@@ -32,7 +32,7 @@ func NewFileResources(file string) ([]FileResource, error) {
 	default:
 		fileInfo, err := os.Stat(file)
 		if err != nil {
-			return nil, fmt.Errorf("Checking file '%s'", file)
+			return nil, fmt.Errorf("Checking file '%s': %s", file, err)
```
praveenrewar: nit: I think we can remove the filename from the error message as the error from os.Stat() always contains the file name.

praveenrewar (approved):

PR #489, opened 2024-06-10
Adding workflow to update go

rohitagg2020:
Adding workflow to update go


praveenrewar (approved):
LGTM!

joaopapereira (changes requested):
In order for this to do a full update we maybe should update our actions to use:
```yaml
    - name: Set Up Go
      uses: actions/setup-go@v5
      with:
        go-version-file: go.mod
        check-latest: true
```
To ensure that we do not need to update the github actions in a separate commit.
Can you please do the change in all the github actions that need it?

rohitagg2020:
> In order for this to do a full update we maybe should update our actions to use:
> 
> ```yaml
>     - name: Set Up Go
>       uses: actions/setup-go@v5
>       with:
>         go-version-file: go.mod
>         check-latest: true
> ```
> 
> To ensure that we do not need to update the github actions in a separate commit. Can you please do the change in all the github actions that need it?

Sounds good

PR #498, opened 2024-07-16
Bump go version to 1.22.5

devacts:
Bump go version to 1.22.5

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module carvel.dev/kbld
 
-go 1.22.3
+go 1.22.5
 
 require (
 	carvel.dev/imgpkg v0.42.2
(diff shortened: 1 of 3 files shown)


praveenrewar (approved):

joaopapereira:
Can we, as part of this PR also change the test that is failing? I am assuming it is just changing to a different image or SHA that exists in docker hub.
Thanks

devacts:
> Can we, as part of this PR also change the test that is failing? I am assuming it is just changing to a different image or SHA that exists in docker hub. Thanks

@joaopapereira It was failing as the image for cloudfoundry/capi-kpack-watcher doesn't exist anymore in dockerhub. So I have removed the check for capi-kpack-watcher image from the relevant tests

Issue #527, opened 2025-02-06
kbld integration with maven-jib


sabeer6870:
**Describe the problem/challenge you have**
We are currently using maven jib to build the docker image for spring boot java application and would like to build it using the kbld from the source. Something very similar to the [existing](https://github.com/carvel-dev/kbld/issues/128) bazel support feature requested here.

**Describe the solution you'd like**
Given a helm chart which uses java application docker image, kbld should be able to build the image from source. 

To build docker image using maven jib, user should be able to write `Config` in the helm chart which will be read by kbld to build the image. Example,

```
sources:
- image: "docker.io/images/myimage:1.0.0"
  path: "src/path_where_mvn_complie_command_will_run"
  jib:
    buildDocker:  
       moduleName: "<module_name>" <---- docker image for this module will be built
destinations:
- image: "docker.io/images/myimage:1.0.0"
  newImage: "local-registry/images/myImage:1.0.0"
  tags: [ "latest", "v1" ]
```

image sha256 of the locally built image can be found at `target/jib-image.digest`, ref : https://github.com/GoogleContainerTools/jib/issues/1443

**Perquisites**
* maven must be installed
* docker must be installed
---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

sabeer6870:
Any update on this? Any ETA?

mamachanko:
> Any update on this? Any ETA?

@sabeer6870 Unfortunately, the core maintainers are lacking the bandwidth right now to implement this feature. You are more than welcome to submit a contribution.

joaopapereira:
Another thing that might help is if you point to a simple app example that whoever picks this issue can use to test.

sabeer6870:
this is still needed, will pick up as soon as I have some cycles.

BulldromeQ:
We have submitted a PR to upstream the Jib build integration! Please see #596 for the implementation.

PR #563, opened 2025-10-22
#527 - Adding structure to support maven jib build

vishrantgupta:
Adding support for maven jib build.

Example:

```
kbld -f examples/simple-app-build-local-mvn/build.yml
```


joaopapereira:
This was finished via #596

PR #585, opened 2026-04-07
Add Buildah support

joaopapereira:
Based on the PR #543 from @Baenimyr  with the following changes:
- Compressed all the commits from @Baenimyr into a single one to make it easier for me to ensure DCO passes
- Made changes to the original workflow because our builders do operations in a different order, where tagging is done afterwards and not by the builder itself

### DIFF
--- a/pkg/kbld/builder/buildah/buildah.go
+++ b/pkg/kbld/builder/buildah/buildah.go
@@ -1,7 +1,7 @@
 // Copyright 2026 The Carvel Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-// Package buildah enable
+// Package buildah provides a builder implementation using Buildah.
 package buildah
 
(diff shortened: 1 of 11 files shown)


pkg/kbld/builder/buildah/buildah.go
```diff
@@ -0,0 +1,191 @@
...
+	if pushErr != nil {
+		return "", pushErr
+	}
+	remoteName := fmt.Sprintf("%s@%s", imgDst.NewImage, digest)
+	prefixedLogger.WriteStr("Image build : %s", remoteName)
+	return remoteName, nil
+}
+
+// Push the buildah manifest and return the digest
+func Push(src string, dest string,
```
praveenrewar, line 131: Do we need to export this function?

PR #596, opened 2026-05-27
feat: Add maven jib build support

BulldromeQ:
## Summary
This pull request introduces support for Maven Jib builds, upstreaming the changes developed in the `vishrantgupta/kbld` fork. 

The changes include:
* Adding structure to support Maven Jib build (`pkg/kbld/builder/maven/jib.go`).
* Updating image building configs to allow Jib integrations.
* Allowing insecure registries with Jib builder.
* Adding a sample app and end-to-end tests for `simple-app-build-local-mvn`.


pkg/kbld/builder/maven/jib.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2024 The Carvel Authors.
```
joaopapereira:
```suggestion
// Copyright 2026 The Carvel Authors.
```

pkg/kbld/builder/maven/jib.go
```diff
@@ -0,0 +1,96 @@
...
+	defer func() {
+		_, _ = prefixedLogger.Write([]byte(
+			"finished build (using kbld jib build)\n"))
+	}()
+
+	tag := opts.Tag
+	if tag == nil {
+		tag = &defaultImageTag
+	}
+	targetImage := fmt.Sprintf("%s:%s", image, *tag)
```
joaopapereira: Nit: I feel like it is more natural not to use pointers here, but just my opinion

```suggestion
	tag := defaultImageTag
	if opts.Tag== nil {
		tag = *opts.Tag
	}
	targetImage := fmt.Sprintf("%s:%s", image, tag)
```

pkg/kbld/config/config_jib.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2024 The Carvel Authors.
```
joaopapereira:
```suggestion
// Copyright 2026 The Carvel Authors.
```

pkg/kbld/image/built.go
joaopapereira:
```suggestion
"path/filepath"

ctlbbz "carvel.dev/kbld/pkg/kbld/builder/bazel"
ctlbdk "carvel.dev/kbld/pkg/kbld/builder/docker"
ctlbko "carvel.dev/kbld/pkg/kbld/builder/ko"
ctlbkb "carvel.dev/kbld/pkg/kbld/builder/kubectlbuildkit"
ctlbmvn "carvel.dev/kbld/pkg/kbld/builder/maven"
ctlbpk "carvel.dev/kbld/pkg/kbld/builder/pack"
ctlconf "carvel.dev/kbld/pkg/kbld/config"
```

pkg/kbld/image/built.go
```diff
@@ -25,13 +26,28 @@ type BuiltImage struct {
 	kubectlBuildkit ctlbkb.KubectlBuildkit
 	ko              ctlbko.Ko
 	bazel           ctlbbz.Bazel
+	jib             maven.Jib
```
joaopapereira:
```suggestion
	jib             ctlbmvn.Jib
```

pkg/kbld/image/built.go
```diff
@@ -25,13 +26,28 @@ type BuiltImage struct {
...
+	Pack            ctlbpk.Pack
+	KubectlBuildkit ctlbkb.KubectlBuildkit
+	Ko              ctlbko.Ko
+	Bazel           ctlbbz.Bazel
+	Jib             maven.Jib
+}
 
-	return BuiltImage{url, buildSource, imgDst, docker, dockerBuildx, pack, kubectlBuildkit, ko, bazel}
+// NewBuiltImage creates a new BuiltImage.
+func NewBuiltImage(url string, buildSource ctlconf.Source,
```
joaopapereira, line 44: This is a breaking change, I like it, and think that we should keep it. Nevertheless, we should let people know in the release notes about it.

pkg/kbld/image/built.go
```diff
@@ -25,13 +26,28 @@ type BuiltImage struct {
...
-	kubectlBuildkit ctlbkb.KubectlBuildkit, ko ctlbko.Ko, bazel ctlbbz.Bazel) BuiltImage {
+// BuildersOpts contains all the builders used to construct a BuiltImage.
+type BuildersOpts struct {
+	Docker          ctlbdk.Docker
+	DockerBuildx    ctlbdk.Buildx
+	Pack            ctlbpk.Pack
+	KubectlBuildkit ctlbkb.KubectlBuildkit
+	Ko              ctlbko.Ko
+	Bazel           ctlbbz.Bazel
+	Jib             maven.Jib
```
joaopapereira:
```suggestion
	Jib             ctlbmvn.Jib
```

pkg/kbld/image/factory.go
joaopapereira:
```suggestion
"fmt"

ctlbbz "carvel.dev/kbld/pkg/kbld/builder/bazel"
ctlbdk "carvel.dev/kbld/pkg/kbld/builder/docker"
ctlbko "carvel.dev/kbld/pkg/kbld/builder/ko"
ctlbkb "carvel.dev/kbld/pkg/kbld/builder/kubectlbuildkit"
ctlbmvn "carvel.dev/kbld/pkg/kbld/builder/maven"
ctlbpk "carvel.dev/kbld/pkg/kbld/builder/pack"
```

pkg/kbld/image/factory.go
```diff
@@ -72,9 +73,19 @@ func (f Factory) New(url string) Image {
 		kubectlBuildkit := ctlbkb.NewKubectlBuildkit(f.logger)
 		ko := ctlbko.NewKo(f.logger)
 		bazel := ctlbbz.NewBazel(docker, f.logger)
+		maven := maven2.NewMavenJib(docker, f.logger)
```
joaopapereira:
```suggestion
		maven := ctlbmvn.NewMavenJib(docker, f.logger)
```

pkg/kbld/builder/maven/jib.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"bytes"
+	ctlbdk "carvel.dev/kbld/pkg/kbld/builder/docker"
+	"carvel.dev/kbld/pkg/kbld/config"
+	ctllog "carvel.dev/kbld/pkg/kbld/logger"
+	"errors"
+	"fmt"
+	"io"
+	"os/exec"
+	"path/filepath"
```
joaopapereira, line 13:
```suggestion
	"bytes"
	"errors"
	"fmt"
	"io"
	"os/exec"
	"path/filepath"
	
	ctlbdk "carvel.dev/kbld/pkg/kbld/builder/docker"
	"carvel.dev/kbld/pkg/kbld/config"
	ctllog "carvel.dev/kbld/pkg/kbld/logger"
```

pkg/kbld/builder/maven/jib.go
```diff
@@ -0,0 +1,96 @@
...
+	_, _ = prefixedLogger.Write([]byte(fmt.Sprintf(
+		"running command: %s\n", cmd)))
+
+	if err := cmd.Run(); err != nil {
+		_, _ = prefixedLogger.Write([]byte(fmt.Sprintf(
+			"error: %s\n", err)))
+		return ctlbdk.TmpRef{}, err
+	}
+
+	inspectData, err := b.docker.Inspect(targetImage)
```
joaopapereira: This is somewhat a race condition where if 2 builds are happening at the same time for the same image and no tags are provided, we do not know what image SHA we will get back. This might be a bit of a corner case, but I am not 100% sure I like this approach.
In other builders, we read the logs of the building tool and parse it. It may be more brittle if the verbiage changes between versions, but at least we know we tag the correct SHA.
Does maven output from jib provide the SHA in the logs?

joaopapereira (changes requested):
The code as a whole looks good, but I would like to see some e2e tests. Take a look at https://github.com/carvel-dev/kbld/blob/develop/test/e2e/build_pack_test.go to make sure we have some assurance that this feature does not break in the future.

Made some form suggestions, mainly around imports and naming, with a NIT that I do not have a strong opinion about; I leave that to you.

My main concern is the fact that we build it and then ask the registry what the SHA is; this can cause race conditions or even malicious behavior.

pkg/kbld/builder/maven/jib.go
```diff
@@ -0,0 +1,96 @@
...
+
+// NewMavenJib creates a new Jib builder.
+func NewMavenJib(docker ctlbdk.Docker, logger ctllog.Logger) Jib {
+	return Jib{docker: docker, logger: logger}
+}
+
+// Run executes the Maven Jib build.
+func (b *Jib) Run(image, directory string,
+	opts config.SourceJibRunOpts) (ctlbdk.TmpRef, error) {
+	prefixedLogger := b.logger.NewPrefixedWriter(image + " | ")
```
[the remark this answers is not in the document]
BulldromeQ: This e2e test will require `mvn` available in the CI runner, I will need more information, we can skip this I believe.

pkg/kbld/builder/maven/jib.go
BulldromeQ: Done.

pkg/kbld/config/config_jib.go
BulldromeQ: Done.

pkg/kbld/builder/maven/jib.go
BulldromeQ: Done.

pkg/kbld/image/built.go
BulldromeQ: Done.

pkg/kbld/image/built.go
BulldromeQ: Done.

pkg/kbld/image/built.go
BulldromeQ: Done.

pkg/kbld/image/factory.go
BulldromeQ: Done.

pkg/kbld/image/factory.go
BulldromeQ: Done.

pkg/kbld/builder/maven/jib.go
BulldromeQ: Yea, agreed — switched to value, defaults to "latest" when no tag is set.

pkg/kbld/image/built.go
BulldromeQ: Agreed, will call it out in the release notes.

pkg/kbld/builder/maven/jib.go
BulldromeQ: The race only hits the latest fallback path — when a tag is set (our consumer always passes a unique one) the inspect can't collide. Ok to leave for now and revisit parsing the SHA from logs as a follow-up?

vg011895 (approved):

pkg/kbld/builder/maven/jib.go
```diff
@@ -0,0 +1,96 @@
...
+			errors.New("Expected target to be specified, but was not")
+	}
+
+	// Base arguments for the Maven Jib command.
+	cmdArgs := []string{
+		"compile",
+		"jib:dockerBuild",
+		"-Dimage=" + targetImage,
+		"-Djib.allowInsecureRegistries=true",
+	}
```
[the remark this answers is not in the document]
BulldromeQ: Our downstream launch tool currently relies on this being enabled by default. We can look into making this configurable in a follow-up coordinated change.

BulldromeQ:
> The code as a whole looks good, but I would like to see some e2e tests. Take a look at https://github.com/carvel-dev/kbld/blob/develop/test/e2e/build_pack_test.go to make sure we have some assurance that this feature does not break in the future.
> 
> Made some form suggestions, mainly around imports and naming, with a NIT that I do not have a strong opinion about; I leave that to you.
> 
> My main concern is the fact that we build it and then ask the registry what the SHA is; this can cause race conditions or even malicious behavior.

Thanks @joaopapereira , could you help me confirm that CI env have maven setup ahead so I could add some tests for these changes.

pkg/kbld/builder/maven/jib.go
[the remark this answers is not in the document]
joaopapereira: The file that needs to be changed to install mvn is https://github.com/carvel-dev/kbld/blob/develop/hack/Dockerfile.dev. Give it a look and see if there is an easy way to add the Java and Maven environment.

pkg/kbld/builder/maven/jib.go
joaopapereira: Today, your consumer does, but if for some reason they stop doing it, we are relying on other people to keep consistency in our tools, which is concerning. In a world where we have multiple teams or eventually multiple agents doing work on a code base at any given time will only increase the possibility of this becoming an issue.
Another option we can use is the same technique we have for docker images building where we generate a temporary tag inside kbld code and then we use that tag, check https://github.com/carvel-dev/kbld/blob/develop/pkg/kbld/builder/docker/docker.go#L63, afterwards we also do the inspect and ensure that the image is tagged with the tag the user asks for.

vg011895: > where if 2 builds are happening at the same time for the same image and no tags are provided

Is this considered a poor build practice, or is this a common industry standard and we should handle that case? @joaopapereira

joaopapereira: This is an issue with the current design. We cannot predict what people will do with kbld, also this can also happen if someone is pushing the same tag to the repository.
What I want to make sure is that we cross our T's and dot our I's, the build already fail for a number of different reasons unrelated to kbld so we should make sure that kbld does behave well. Also this is not such a big change in the current code that we can do to ensure no randomess happens when building

pkg/kbld/builder/maven/jib.go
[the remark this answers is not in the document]
BulldromeQ: Done! Added `default-jdk-headless` and `maven` to `hack/Dockerfile.dev`. I also added a full E2E test suite in `test/e2e/build_maven_jib_test.go` and verified that Jib image building compiles and works successfully inside this environment. To completely resolve the race condition, I also updated the builder to build to a unique random temporary tag first before retagging.

pkg/kbld/builder/maven/jib.go
BulldromeQ: I went ahead and implemented the robust solution for this to completely eliminate any possible race condition: we now generate a unique, random temporary tag (via `ctlb.TagBuilder{}`) to run the maven build against, inspect that specific temporary tag, and then retag it to the desired user tag and stable/digest references.

BulldromeQ:
@joaopapereira Update on the recent CI failures in the `test-gh` job (`TestPackBuildAndPushSuccessful` getting a 404 for `jvmkill`): I initially thought this was an external infrastructure flake, but it turns out it was indirectly caused by this PR!

By placing the `pom.xml` inside `test/e2e/assets/simple-app/helloworld`, the `pack` builder in the older E2E tests (which targets the parent `simple-app` dir) mistakenly detected the project as a Java application. This triggered the legacy Cloud Foundry OpenJDK Buildpack, which then failed because of its broken download link for `jvmkill`.

I've just pushed a commit that moves the Java app into its own dedicated directory at `test/e2e/assets/maven-app`. This cleanly separates the Go and Java assets and completely fixes the CI issue.

joaopapereira:
I was looking at cnb in docker hub and the latest sha is sha256:17ea21162ba8c7717d3ead3ee3836a368aced7f02f2e59658e52029bd6d149e7 not sure if that will be enough but they did not update the image for 3 years. can you check with that image

BulldromeQ:
@joaopapereira I've updated the pack builder image to the one you suggested (`index.docker.io/cloudfoundry/cnb@sha256:17ea21162ba8c7717d3ead3ee3836a368aced7f02f2e59658e52029bd6d149e7`) in the e2e tests. Also, I've moved the java app into its own directory (`test/e2e/assets/maven-app`) so it won't conflict with the go app in `simple-app`. Let's see if the CI turns green now!

BulldromeQ:
@joaopapereira The suggested builder image (`17ea21162ba8c7717d3ead3ee3836a368aced7f02f2e59658e52029bd6d149e7`) failed during detection with: `ERROR: validating stack mixins: buildpack paketo-buildpacks/apache-tomcat@7.13.7 does not support stack io.buildpacks.stacks.bionic`.

Because we've already fixed the root cause of the previous 404 error (by moving `maven-app` completely out of the `simple-app` folder so the older test no longer misidentifies it as a Java app), the original builder image (`83270cf59e8944be0c544e45fd45a5a1f4526d7936d488d2de8937730341618d`) should work perfectly again! 

I've reverted the image SHA back to the original in the test. The CI should hopefully pass cleanly now.

joaopapereira (approved):
LGTM

joaopapereira:
@BulldromeQ, thanks for the long feedback loop here. I am happy that we got to the current point.
We need to update https://github.com/carvel-dev/carvel/tree/develop/site/content/kbld/docs/develop to include the new configuration for Jib before we release a new kbld version. Can you please do that PR?

BulldromeQ:
@joaopapereira Thanks! I've just opened the PR with the new Jib configuration documentation in the carvel repo here: https://github.com/carvel-dev/carvel/pull/828.

PR #598, opened 2026-06-16
Replace SHA1 usage with SHA256 for enhanced security

nikhilsagotiya:
**Summary**
Replaces crypto/sha1 with crypto/sha256 in the e2e packaging test to eliminate SHA-1 usage.

**What changed**

- Removed crypto/sha1 import from test/e2e/packaging_test.go
- Renamed sha1File helper to sha256File, updated its implementation to use crypto/sha256
- Replaced the hardcoded SHA-1 expected value (ca02098b7f6cee61836119233e3b43d1a92083b4) with its SHA-256 equivalent (e084f139ad114ef6c56dc7343f7d6bd8d2cd11a16e77e00aca97b5e5d18c3a56), captured from a live CI run

Why a hardcoded SHA-256 (not just cross-run comparison)
The SHA-256 value was captured from a CI run where all source images are pinned by @sha256: digest, making the tar output fully deterministic. Keeping the hardcoded expected value preserves both:

1. Concurrency-determinism — same output regardless of --concurrency level
2. Regression detection — catches unexpected changes to the packaging format

**Security motivation**
SHA-1 is not FIPS-approved. Although this test file is excluded from production binaries via //go:build e2e, removing the crypto/sha1 import eliminates the cryptographic audit finding entirely.

**Testing**
This change only affects the e2e test file (//go:build e2e). No production code is modified. The test passes in CI with the hardcoded SHA-256 value.

### DIFF
--- a/test/e2e/packaging_test.go
+++ b/test/e2e/packaging_test.go
@@ -6,7 +6,7 @@
 package e2e
 
 import (
-	"crypto/sha1"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -253,7 +253,7 @@ overrides:
   preresolved: true
 `
 
-	expectedPackagedSHA := "ca02098b7f6cee61836119233e3b43d1a92083b4"
+	expectedPackagedSHA256 := "e084f139ad114ef6c56dc7343f7d6bd8d2cd11a16e77e00aca97b5e5d18c3a56"
 
 	path := "/tmp/kbld-test-pkg-unpkg-successful-with-many-images"
 	defer os.RemoveAll(path)
@@ -262,21 +262,21 @@ overrides:
 		StdinReader: strings.NewReader(input),
 	})
 
-	actualSHA := sha1File(t, path)
+	actualSHA256 := sha256File(t, path)
 
-	// Assert that concurrently writing to tar doesn't affect sha
-	if actualSHA != expectedPackagedSHA {
-		t.Fatalf("Expected package sha to be same >>>%s<<< to match >>>%s<<<", actualSHA, expectedPackagedSHA)
+	// Assert that concurrently writing to tar doesn't affect sha256
+	if actualSHA256 != expectedPackagedSHA256 {
+		t.Fatalf("Expected package sha256 to be same >>>%s<<< to match >>>%s<<<", actualSHA256, expectedPackagedSHA256)
 	}
 
 	kbld.RunWithOpts([]string{"package", "-f", "-", "--output", path, "--concurrency=5"}, RunOpts{
 		StdinReader: strings.NewReader(input),
 	})
 
-	actualSHA = sha1File(t, path)
+	actualSHA256 = sha256File(t, path)
 
-	if actualSHA != expectedPackagedSHA {
-		t.Fatalf("Expected package sha to be same >>>%s<<< to match >>>%s<<<", actualSHA, expectedPackagedSHA)
+	if actualSHA256 != expectedPackagedSHA256 {
+		t.Fatalf("Expected package sha256 to be same >>>%s<<< to match >>>%s<<<", actualSHA256, expectedPackagedSHA256)
 	}
 
 	kbld.RunWithOpts([]string{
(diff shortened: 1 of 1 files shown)


test/e2e/packaging_test.go
```diff
@@ -253,30 +253,23 @@ overrides:
...
 		StdinReader: strings.NewReader(input),
 	})
 
-	actualSHA := sha1File(t, path)
-
-	// Assert that concurrently writing to tar doesn't affect sha
-	if actualSHA != expectedPackagedSHA {
-		t.Fatalf("Expected package sha to be same >>>%s<<< to match >>>%s<<<", actualSHA, expectedPackagedSHA)
-	}
+	hash1 := sha256File(t, path)
```
aroradaman: nit: Makes more sense to have `hash256`.

test/e2e/packaging_test.go
```diff
@@ -253,30 +253,23 @@ overrides:
   preresolved: true
 `
 
-	expectedPackagedSHA := "ca02098b7f6cee61836119233e3b43d1a92083b4"
```
aroradaman, line 256: Can we check if we actually need this SHA, if required can we have a 256 SHA digest over here?

test/e2e/packaging_test.go
nikhilsagotiya: Renamed hash1 → hash256 and hash2 → hash256C5 to make it clear both variables hold SHA-256 digests and which concurrency run each corresponds to.

test/e2e/packaging_test.go
nikhilsagotiya: The hardcoded SHA-1 was serving two purposes: (1) concurrency-determinism, and (2) byte-exact regression detection. 
We've preserved (1) with the direct cross-run comparison. For (2), the kbld unpackage step that follows provides a stronger functional assertion — if the tar format regresses, unpacking will fail. A hardcoded SHA-256 baseline would also need to be re-computed every time an image digest in the test input is updated, which adds maintenance overhead.

praveenrewar: > A hardcoded SHA-256 baseline would also need to be re-computed every time an image digest in the test input is updated, which adds maintenance overhead.

I agree, however, I think we should make that decision in a separate PR (if we want to) and for now just update sha1 to sha256 to avoid any regressions later on.

nikhilsagotiya: Done — captured the SHA-256 value (e084f139...) from the CI run and hardcoded it, restoring the original test pattern with SHA-256 in place of SHA-1.

nikhilsagotiya:
> ## Pull request overview
> This PR updates the e2e packaging determinism test to avoid SHA-1 usage by switching file hashing to SHA-256.
> 
> **Changes:**
> 
> * Replaced `crypto/sha1` with `crypto/sha256` in `test/e2e/packaging_test.go`.
> * Renamed the helper from `sha1File` to `sha256File` and updated the hashing implementation accordingly.
> * Updated the determinism assertion logic around `--concurrency` packaging runs (but see review comment re: hardcoded baseline vs. hash-to-hash comparison).
> 
> Comments suppressed due to low confidence (1)
> **test/e2e/packaging_test.go:260**
> 
> * The PR description says the hardcoded expected hash was removed and the test now compares the SHA-256 from `--concurrency=1` against the SHA-256 from `--concurrency=5`. In this hunk the test still uses a hardcoded SHA-256 (`expectedPackagedSHA256`) and checks both runs against it, which is both a mismatch with the PR description and more brittle across environments/tar metadata changes.
> 
> ```
> 	expectedPackagedSHA256 := "e084f139ad114ef6c56dc7343f7d6bd8d2cd11a16e77e00aca97b5e5d18c3a56"
> 
> 	path := "/tmp/kbld-test-pkg-unpkg-successful-with-many-images"
> 	defer os.RemoveAll(path)
> ```
> 
> 💡 [Add Copilot custom instructions](/carvel-dev/kbld/new/develop?filename=.github/instructions/*.instructions.md) for smarter, more guided reviews. [Learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

The PR description has been updated to accurately reflect

praveenrewar (approved):

PR #600, opened 2026-07-20
Fix error output in JSON mode

### LINKED ISSUE #455: kbld with flag --json swallows error
**What steps did you take:**
I have config.yaml
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kbld-nginx
spec:
  selector:
    matchLabels:
      app: kbld-nginx
  template:
    metadata:
      labels:
        app: kbld-nginx
    spec:
      containers:
      - name: my-app
        image: nginxix
```
And it is throwing error when run kbld command without `--json` flag:
```
$ kbld -f nginx/config.yml 
kbld: Error: 
- Resolving image 'nginxix': GET https://index.docker.io/v2/library/nginxix/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/nginxix Type:repository]]
```

While it will exit without printing any error message or any log if will run with flag `--json`

**What happened:**
When we run kbld command with flag `--json` it is exiting without printing any error message or any log in case of failure.

**What did you expect:**
When we run kbld command with flag `--json` it should print error message  in case of failure.

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

**Environment:**

- kbld version (use `kbld --version`):
- Docker registry used (e.g. `Docker HUB`): 
- OS (e.g. from `/etc/os-release`):

If building images using `docker`:
- Docker CLI Version (e.g. `docker version`)

If building images using `buildkit`:
- Buildkit version (e.g. `kubectl buildkit version`):
- Kubernetes version (e.g. `kubectl version`):

If building images using `pack`
- Pack Version (e.g. `pack version`):

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

snowyukitty:
Fixes #455.

When `--json` is enabled, fatal errors are buffered by the JSON UI. The process then calls `os.Exit(1)`, which skips the deferred flush and leaves users with a non-zero exit code but no diagnostic. Flushing the buffer on this path would put the error on stdout, while JSON failures are expected to keep stdout empty and report a plain-text error on stderr.

This change keeps the parsed UI options available in `main` and bypasses the buffered UI only for JSON-mode failures. Non-JSON errors still use the configured UI, preserving color and TTY behavior; successful JSON output is unchanged.

An end-to-end regression test verifies exit code 1, empty stdout, and a plain-text stderr diagnostic.


cmd/kbld/kbld_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2026 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
```
joaopapereira: I feel like this looks more like and e2e test.  Can you please move it there and I don't think we need this much setup.

joaopapereira (changes requested):
I am going to allow the tests to run, nevertheless I left a comment but in general the code looks ok

cmd/kbld/kbld_test.go
snowyukitty: Moved the regression into the existing invalid-digest e2e case in `test/e2e/resolve_test.go`. It now reuses `BuildEnv` and `Kbld.RunWithOpts`, so the subprocess, environment-variable, and temp-file setup in `cmd/kbld/kbld_test.go` is gone. The focused e2e test, full unit suite, and golangci-lint v2.12.2 pass.

PR #602, opened 2026-09-22
Fix registry hostname case sensitivity, CI runtime pin, and error-message formatting

sameerforge:
## Summary

Comprehensive fix for case-sensitivity handling in image registry operations:

- **Registry hostname comparison**: Made case-insensitive in `EnvKeychain.Resolve` (RFC 1035/1123 DNS names are case-insensitive). Environment variable credentials like `KBLD_REGISTRY_HOSTNAME_*` that differ only in case from the target image's registry host now correctly resolve instead of silently falling back to anonymous auth.

- **Repository path case preservation**: `lowerCaseRepository` now properly handles case-sensitive image reference components. Only the registry hostname is lowercased; repository paths and tags preserve original case, matching behavior of registries like GHCR and Quay.

- **CI reliability**: Added explicit `--container-runtime=docker` to `minikube start` to prevent `docker.service is masked` failures on newer kicbase images.

- **Code quality**: Fixed error message formatting and resolved all golangci-lint violations.

### `pkg/kbld/registry/env_keychain.go`
- Use `strings.EqualFold()` for case-insensitive registry hostname comparison in credential resolution

### `pkg/kbld/builder/docker/docker.go`
Refactored `lowerCaseRepository` function:
- Parse references using `regname.ParseReference()` for proper component extraction
- Only lowercase the registry hostname (case-insensitive per RFC 1035/1123)
- Preserve case-sensitive repository paths and tags
- Fallback to manual parsing if reference parsing fails
- Created named constants (`noSlashIdx`, `registryStart`) to replace magic numbers
- Optimized line lengths to pass golangci-lint 80-character limit

### `pkg/kbld/builder/docker/docker_test.go` (new)
Comprehensive test coverage for `lowerCaseRepository`:
- 7 test cases covering various image reference formats
- Tests verify: mixed-case registries, registry+path combinations, digest references, tags, localhost registries
- Each test ensures only registry hostname is lowercased while paths/tags retain original case

### `.github/workflows/test-gh.yml`
- Added `--container-runtime=docker` flag to minikube start command

### `pkg/kbld/config/config.go`
- Fixed error message formatting (removed doubled space)
- Reformatted line to stay within line-length limits

### Unit Tests
- **`pkg/kbld/registry/env_keychain_test.go`**:
  - `TestEnvKeychainResolveIsCaseInsensitive` — sets `KBLD_TEST_HOSTNAME=MyRegistry.Example.com` and resolves against `myregistry.example.com`, asserting the configured credentials are returned
  - `TestEnvKeychainResolveNoMatchIsAnonymous` — asserts an unset keychain falls back to anonymous credentials (regression guard)

- **`pkg/kbld/builder/docker/docker_test.go`**:
  - `TestLowerCaseRepository` — 7 test cases covering:
    - Registry hostname lowercasing with case-sensitive repository path preservation
    - Mixed case registries and repositories
    - Digest references (with `@sha256:...`)
    - Tag preservation (with `:tag` format)
    - Localhost registries with port numbers
  - All cases verify only registry hostname is lowercased

### CI/Lint Verification
- ✅ All golangci-lint checks pass (0 issues)
- ✅ All unit tests pass
- ✅ CI workflow (`test-gh`) now completes `minikube start` without `docker.service is masked` errors

## Related Issues

This fix aligns kbld with the same case-sensitivity issues addressed in:
- `carvel-dev/kapp-controller` — registry hostname case-insensitive comparison
- `carvel-dev/imgpkg#795` — case-sensitive repository path preservation


pkg/kbld/builder/docker/docker.go
```diff
@@ -340,26 +340,39 @@ func (d Docker) Inspect(ref string) (InspectData, error) {
...
+		registryStart = 0
+	)
+
+	// Attempt to parse reference to properly identify registry vs repository.
+	parsedRef, err := regname.ParseReference(ref)
+	if err == nil {
+		registryStr := strings.ToLower(parsedRef.Context().RegistryStr())
+		registryPart := parsedRef.Context().RegistryStr()
+		registryIdx := strings.Index(ref, registryPart)
+
```
himsngh: Issue flagged by Claude, 

```
docker.go lowerCaseRepository — 🔴 real regression, verified by running it

I cloned the PR branch and tested it directly. All 8 tests the author added pass, but they only cover refs that already contain a slash. I added three cases for the very common "implicit docker.io registry" form and they fail:

lowerCaseRepository("MyImage:Tag")      = "myimage:tag"       want "MyImage:Tag"
lowerCaseRepository("MyImage")          = "myimage"           want "MyImage"
lowerCaseRepository("MyOrg/MyImage:Tag")= "myorg/MyImage:Tag" want "MyOrg/MyImage:Tag" (this one's actually fine by luck)

Root cause: when ref has no explicit registry host, regname.ParseReference fills in a default (index.docker.io), but that literal string doesn't appear in the original ref, so strings.Index(ref, registryPart) returns -1, the primary branch's registryIdx == registryStart check fails, and it falls through to the "fallback" path. The fallback treats whatever text precedes the first / as if it were a registry hostname:
- If there's no / at all (bare image name like MyImage:Tag), it does strings.ToLower(ref) — lowercasing the whole thing, tag included. This directly contradicts the PR's own stated goal ("repository paths and tags preserve original case") and is a real behavior regression for any single-segment reference with mixed-case name/tag.
- If there is a / (e.g. MyOrg/MyImage:Tag), it happens to only lowercase the first segment, which is usually harmless since Docker Hub namespaces are case-insensitive too — but this is coincidental, not by design, and doesn't generalize (e.g. it will also mis-fire on refs with no registry but multiple path segments in some registry configs).

This is untested territory precisely because none of the 7 new test cases use a ref without an explicit host. I'd flag this to the author as needing:
1. A test for the bare name:tag / bare name case.
2. A fix — the cleanest approach is probably to reconstruct from parsedRef.Context().RegistryStr() + parsedRef.Context().RepositoryStr() + tag/digest directly (all obtainable from parsedRef) instead of re-searching for the registry substring in the raw string and falling back to first-slash heuristics.
```

can you take a look?

sameerforge: I've fixed it by detecting implicit vs. explicit registries—preserving case when the registry isn't in the original string, and we added test coverage for all three cases you flagged.

pkg/kbld/builder/docker/docker.go
himsngh: imageDst reaching this branch with an ErrBadName may mean it has uppercase in the repository path (host casing was never rejected by this library).

For GHRC/Quay, a mixed-case repository path like MyOrg/MyApp isn't just "a different destination than lowercase," it's not a valid reference at all; only myorg/myapp can ever exist.

ref: https://github.com/orgs/community/discussions/27086

So with this change, we will be throwing error here if there is an Uppercase repository name. Previously, we were silently updating it to lowercase.

himsngh: The unit tests added here, if we run it in the caller function it will through error (which was not the case before). 

I believe, even before it was a bug (by lowercasing the repository) which this PR addresses (by throwing error). Want to get other's opinion here before approving.

himsngh: cc: @aroradaman / @praveenrewar / @joaopapereira

